Issue #10, opened 2018-09-28
Would it be possible to use an handle instead of a context file ?


GaelGuegan:
All is in the title.

I would like to use the engine with a persistent handle already created in the TPM.
This, in order to do a rsa encryption.

Thank you

AndreasFuchsSIT:
Possible, yes.
Appreciated, yes.
If you want to give it a try, I'll happily help.

I guess one would have to extend only one function: https://github.com/tpm2-software/tpm2-tss-engine/blob/master/src/tpm2-tss-engine.c#L149

E.g. if the keyid starts with 0x it's gonna be interpreted as TPM-persistent.

Patches welcome.

GaelGuegan:
Okay, thanks for the help.
I will see what i can do.

GaelGuegan:
Okay I have done a pull request to correct the issue.

PR #14, opened 2018-10-03
LOADKEY Function: Add management of persistent handle

GaelGuegan:
Here the patch of the issue #10 .

### DIFF
--- a/include/tpm2-tss-engine.h
+++ b/include/tpm2-tss-engine.h
@@ -34,10 +34,15 @@
 #include <openssl/engine.h>
 #include <tss2/tss2_tpm2_types.h>
 
+typedef enum { blob, handle } KEY_TYPE ;
 typedef struct {
     TPM2B_DIGEST userauth;
     TPM2B_PUBLIC pub;
-    TPM2B_PRIVATE priv;
+    union {
+      TPM2B_PRIVATE priv;
+      TPM2_HANDLE handle;
+    };
+    KEY_TYPE privatetype;
 } TPM2_DATA;
 
 #define TPM2TSS_SET_OWNERAUTH ENGINE_CMD_BASE
--- a/include/tpm2-tss-engine.h
+++ b/include/tpm2-tss-engine.h
@@ -34,7 +34,11 @@
 #include <openssl/engine.h>
 #include <tss2/tss2_tpm2_types.h>
 
-typedef enum { blob, handle } KEY_TYPE ;
+typedef enum {
+    KEY_TYPE_BLOB,
+    KEY_TYPE_HANDLE
+} KEY_TYPE;
+
 typedef struct {
     TPM2B_DIGEST userauth;
     TPM2B_PUBLIC pub;
--- a/src/tpm2-tss-engine-rsa.c
+++ b/src/tpm2-tss-engine-rsa.c
@@ -157,7 +157,7 @@ rsa_priv_enc(int flen,
     DBG("Padded digest data (size=%i):\n", digest.size);
     DBGBUF(&digest.buffer[0], digest.size);
 
-    if (tpm2Data->privatetype == handle) {
+    if (tpm2Data->privatetype == KEY_TYPE_HANDLE) {
       r = Esys_Initialize(&ectx, NULL, NULL);
       ERRchktss(rsa_priv_enc, r, goto out);
       r = Esys_Startup(ectx, TPM2_SU_CLEAR);
@@ -192,7 +192,7 @@ error:
 out:
     free(sig);
     if (keyHandle != ESYS_TR_NONE) {
-      if (tpm2Data->privatetype == handle) {
+      if (tpm2Data->privatetype == KEY_TYPE_HANDLE) {
         Esys_TR_Close(ectx, &keyHandle);
       } else {
         Esys_FlushContext(ectx, keyHandle);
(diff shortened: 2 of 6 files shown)


include/tpm2-tss-engine.h
```diff
@@ -34,10 +34,15 @@
 #include <openssl/engine.h>
 #include <tss2/tss2_tpm2_types.h>
 
+typedef enum { blob, handle } KEY_TYPE ;
```
AndreasFuchsSIT: Please namespace the enum members, preferably with the enum typename, i.e. KEY_TYPE_BLOB

GaelGuegan: Okay can I use tpm_tools scripts in my test, in order to generate a persistent handle ?

AndreasFuchsSIT:
Sure, that would even be good for people to look up.
I can also help with the .travis.yml scripts for installing the tpm2-tools on the CI-server

GaelGuegan:
Hey @AndreasFuchsSIT, I try several things but I have still errors because of package dependencies with tpm-tools. I try to add libcurl3-dev & libtool to travis.yml, without success ...

AndreasFuchsSIT:
maybe try requiring libcurl4-openssl-dev isntead of libcurl3-dev

GaelGuegan:
Okay, I have tried what you suggest without success. A new error appear and I changed libgcrypt20-dev to libgcrypt-dev, but now it is the TSS compilation who does not work ... :'(
I am loosing hope.

AndreasFuchsSIT:
Yes, tpm2-tss requires libgcrypt20-dev...
I guess is could be cloned and installed manually, similar to how tpm2-tss is cloned and installed as well.
Then this might just work...

If all fails, we'll repleace the tpm2-tools with some special binary using esapi directly. But let's see first...
________________________________

src/tpm2-tss-engine.c
```diff
@@ -165,16 +166,50 @@ loadkey(ENGINE *e, const char *key_id, UI_METHOD *ui, void *cb_data)
...
 
 	DBG("Loading private key %s\n", key_id);
-
-    if (!tpm2tss_tpm2data_read(key_id, &tpm2Data)) {
+    tpm2Data = OPENSSL_malloc(sizeof(*tpm2Data));
+    if (tpm2Data == NULL) {
+      ERR(tpm2tss_tpm2data_read, ERR_R_MALLOC_FAILURE);
+      goto error;
+    }
+    if (strncmp(key_id, "0x", 2) == 0) {
```
dwmw2: Please let's standardise a URI format for this, perhaps based on https://tools.ietf.org/html/draft-mavrogiannopoulos-tpmuri-01

src/tpm2-tss-engine-rsa.c
```diff
@@ -246,8 +261,19 @@ rsa_priv_dec(int flen,
...
-    r = init_tpm_key(&ectx, &keyHandle, tpm2Data);
-    ERRchktss(rsa_priv_dec, r, goto out);
+    if (tpm2Data->privatetype == handle) {
+      r = Esys_Initialize(&ectx, NULL, NULL);
+      ERRchktss(rsa_priv_dec, r, goto out);
+      r = Esys_Startup(ectx, TPM2_SU_CLEAR);
+      ERRchktss(rsa_priv_dec, r, goto out);
+      r = Esys_TR_FromTPMPublic(ectx, tpm2Data->handle, ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, &keyHandle);
+      ERRchktss(rsa_priv_dec, r, goto out);
+      r = Esys_TR_SetAuth(ectx, ESYS_TR_RH_OWNER, &tpm2Data->userauth);
```
dwmw2: You want to do a ReadPublic call here and check the NODA flag, then if the NODA flag isn't set, use a UI to prompt for a password if you don't already have one.

See http://git.infradead.org/users/dwmw2/openconnect.git/blob/tpm2:/gnutls_tpm2_esys.c#l277 for example.

src/tpm2-tss-engine.c
AndreasFuchsSIT: That thing is outdated for TPM2...
No uuids anymore and "storage" has different means of system vs user...
I'd wanna go with 0x or 0x81 for the moment, until we have figured out a tpm2uri scheme or adapted tpmuri...

src/tpm2-tss-engine-rsa.c
GaelGuegan: Okay I corrected it

dwmw2:
For the password handling part, please do some dev testing of this manually with the engine being used from OpenConnect. You don't need a VPN; just point it at any HTTPS server and you'll test the password prompting.

For the builds with GnuTLS, it will automatically prompt the user as needed for the owner hierarchy password or the parent key password as appropriate, and for a password for the key itself.

If NODA is set on a key or a persistent parent (and always for a hierarchy password), it'll try the empty password and then prompt the user if that fails.

I have done my testing with all variants of hierarchy password set/unset, parent key password set/unset, actual key password set/unset. Most of my tests fail with the ENGINE while they work in the GnuTLS build.

GaelGuegan:
I have revert to one of first commit.
If you want to correct other issue please make a pull request.

I cannot correct everyting ... sorry

AndreasFuchsSIT:
That's fine, I'll take a stab.
Thanks for the effort up to here !

AndreasFuchsSIT:
I took your stuff an refactored a bit. Please see #33 and reviews are welcome.
Closing this PR for the moment.

PR #22, opened 2018-10-12
Support standardised PEM format

dwmw2:


src/tpm2-tss-engine-rsa.c
```diff
@@ -290,6 +290,10 @@ populate_rsa(RSA *rsa) {
     if (tpm2Data == NULL)
         goto error;
 
+    /* Default exponent, if it's stored as zero. */
```
AndreasFuchsSIT: I don't like this for 2 reasons:
- The populate_rsa method should not change the tpmkey object. It should've been declared as const in Line 289 I guess.
- We should not change the public area to be loaded by the TPM at all. If the TPM gives us a TPM2B_PUBLIC with exponent = 0 then we keep it as 0... Everything else may lead to problems.

My proposal: Change line 311:
`BN_set_word(rsa->e, tpm2Data->pub.publicArea.parameters.rsaDetail.exponent ? tpm2Data->pub.publicArea.parameters.rsaDetail.exponent : 0x10001);`

AndreasFuchsSIT:
Thanks !

Issue #13, opened 2018-10-02
Asked for password when none required


dwmw2:
I created a fake X.509 cert to go with a key generated with tpm2tss-genkey. Now the `openssl s_client` invocation like the one in #12 works fine, except of course that I can't actually authenticate to my VPN server. It does offer the fake client cert correctly though.

However, I'm asked for a password even though I didn't specify one when creating the key.

AndreasFuchsSIT:
Yes, that is correct.
The TPM actually has no notion of "unset password" but merely of empty password.
So this engine exposes exactly that behavior.

dwmw2:
Can we not try with an empty password and prompt only if that doesn't work?

The storage format of openssl_tpm2_engine includes an indication that the password is empty.

AndreasFuchsSIT:
Change required here: https://github.com/tpm2-software/tpm2-tss-engine/blob/master/src/tpm2-tss-engine.c#L174

dwmw2:
You'll note you now have the 'emptyAuth' hint from the PEM file itself, but I haven't wired that up for you. (Neither have I wired up the 'parent' field, except to reject anything but RH_OWNER for now).

dwmw2:
For reference, see also the various `request_passphrase()` calls in my code in
http://git.infradead.org/users/dwmw2/openconnect.git/blob/tpm2:/gnutls_tpm2_esys.c where we ask _(repeatedly, if needs be)_ for passwords for the hierarchy, the parent key, and the wrapped key itself.

Although the header there says LGPLv2.1 it's all my code (except the bits I copied from you) and you're welcome to use it under the 3-clause BSD licence if any of it is useful, or if you just want to be careful about pollution from looking.

AndreasFuchsSIT:
I noticed, just did not want for this to be forgotten.
I have an ENGINE_CTRL for ownerpass, thus I don't want to call the user repeatedly, but thanks!

dwmw2:
Likewise, OpenConnect will load the engine but will expect UI callbacks if the engine wants anything; it doesn't make special arrangements (and how could it? It doesn't even know you *have* a password on the hierarchy).

dwmw2:
> Given this, do we maybe want to add a "parent-noauth" flag to the PEM-format ?

That seems like a sane suggestion. @jejb?

dwmw2:
[...]

PR #35, opened 2018-10-18
Add option for password-less operations

AndreasFuchsSIT:
Addresses: #13


AndreasFuchsSIT:
@safayetahmedatge if you could have a look at this. It should solve your request.
Also look at newly added integration test.

PR #40, opened 2018-10-25
Switch to final assigned OIDs for key types

dwmw2:
Turns out I was right to be paranoid about chasing up the informally
assigned OIDs we were using. Those have been designated as non-leaf nodes
now, and we have new ones.

Signed-off-by: David Woodhouse <dwmw2@infradead.org>


dwmw2:
It's on a spreadsheet on Monty's laptop so far. Will update when we have a public reference.

dwmw2:
We also need to work on the public definition of the ASN.1 format too, and the OIDs will be present in that.

dwmw2:
We will work on the public documentation including the OIDs but please could we merge this fix in the meantime for compatibility.

PR #44, opened 2018-11-08
doc: Add missing build steps for tpm2-tss

jhachenbergerSIT:
The build steps for the tpm2-tss dependency are incomplete (they miss
the two calls: ./bootstrap and ./configure, see [1])

[1] https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md

Signed-off-by: Julien Hachenberger <julien.hachenberger@sit.fraunhofer.de>

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,6 +25,8 @@ sudo apt -y install \
   ruby-ronn
 git clone --depth=1 http://www.github.com/tpm2-software/tpm2-tss
 cd tpm2-tss
+./bootstrap
+./configure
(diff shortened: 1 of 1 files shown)


jhachenbergerSIT:
Instead of adding those two lines, we could also completely get rid of the build steps for the tpm2-tss project and link to their [INSTALL](https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md) instruction instead. This would also fix the problem of the missing dependency which occur while building the tpm2-tss project with the here mentioned steps.

PR #55, opened 2018-11-26
Enable emptyAuth detection for noda persistent keys

AndreasFuchsSIT:
If a persistent key is requested and this key has the noda flag set, we will attempt to authenticate with an
empty auth. This helps to save the user a password prompt.

Fixes: #13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ addons:
     - libssl-dev
     - uthash-dev
     - pandoc
+    - expect
 
 install:
   - git clean -xdf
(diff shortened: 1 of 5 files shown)


JuergenReppSIT:
((r & ~TSS2_RC_N_MASK == TPM2_RC_BAD_AUTH)  would be better than  (r == 0x9a2) .
and
((r & ~TSS2_RC_LAYER_MASK) ==  TSS2_BASE_RC_GENERAL_FAILURE)  would be better than  
(r == 0x00070001).

AndreasFuchsSIT:
Changed

PR #61, opened 2018-12-20
tests: use tpm2-tools commands from released version instead of master

martinezjavier:
The tpm2-tss-engine CI pipeline clones the tpm2-tss using the latest tag
(2.1.0) but uses the master branch for the tpm2-tools. Instead let's use
the latest release tag for the tpm2-tools as well so we test against a
stable baseline.

This will also make it easier for users to build tpm2-tss-engine since
can use the tpm2-tss and tpm2-tools packages provided by their distro.

Otherwise will need to build the tpm2-tools from source if they want to
run the tests (i.e: with the make check target).

Most of the changes are just about using different command options but 2
are more significant:

* tpm2_evictcontrol tool requires a persistent handle since since is
  still not able to get the first available vacant persistent handle.

* tpm2_flushcontext is not present in the latest release of the tools
  so contexts can't be explicitly flushed. I don't think this is a big
  issue since the simulator is started just before executing the tests.

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,7 +81,7 @@ install:
   - rm ${PWD}/../installdir/usr/local/lib/*.la
   - popd
 # tpm2-tools
-  - git clone --depth=1 https://github.com/tpm2-software/tpm2-tools.git
+  - git clone --depth=1 -b 3.X https://github.com/tpm2-software/tpm2-tools.git
   - pushd tpm2-tools
   - mkdir m4 || true
   - cp ../autoconf-archive-2017.09.28/m4/ax_code_coverage.m4 m4/
--- a/test/rsasign_parent.sh
+++ b/test/rsasign_parent.sh
@@ -6,6 +6,8 @@ export OPENSSL_ENGINES=${PWD}/.libs
 export LD_LIBRARY_PATH=$OPENSSL_ENGINES:${LD_LIBRARY_PATH-}
 export PATH=${PWD}:${PATH}
 
+PHANDLE=0x81010001
+
 DIR=$(mktemp -d)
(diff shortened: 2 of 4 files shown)


martinezjavier:
@AndreasFuchsSIT this is more of an RFC pull-request since I don't know if you prefer to be tested against tpm2-tools master.

I just started looking at your tpm2-tss-engine project and made this change locally, so I thought that could share with you.

AndreasFuchsSIT:
Thanks for the input.
I guess it was best to do both via the config matrix.

However currently it's failing. Have you already looked into why that is ?

AndreasFuchsSIT:
@martinezjavier Have you made any progress with this ?

AndreasFuchsSIT:
@martinezjavier So the 3.X series of the tpm2-tools does not have a flushcontext.
Locally you probably have the program as a left-over from a master-test (that was my thing here as well).
But when building from scratch, that's why we fail. No tab/rm and no manual flush => tpm full...

martinezjavier:
@AndreasFuchsSIT sorry I didn't have time to make any progress on this.

> Locally you probably have the program as a left-over from a master-test (that was my thing here as well)

Right, locally I think that used the tpm2-abrmd so I didn't hit the full TPM. So what do you suggest? Probably we should just drop this PR then since the tools are only needed for the tests.

AndreasFuchsSIT:
I guess, I'll close this PR and will checkout a certain commit of tpm2-tools until 4.X is ready.
But I found a session-leak in combination with tpm2-tss 2.0 and 2.1 that I'll try to fix. It's fixed with tpm2-tss master/2.2 though...

PR #64, opened 2019-01-05
Free memory allocated for RSA key components

raghuncstate:
Code fails to free partially allocated components of the RSA key in the
error path.

If a BIGNUM other than the first one(BN *n) fails to allocate memory,
the error path does not free allocated memory of the other components.
This issue occurs only for OpenSSL versions > 1.1.x

Signed-off-by: Raghu Krishnamurthy <raghu.ncstate@icloud.com>


raghuncstate:
Done.

Issue #31, opened 2018-10-17
Request use of SHA256 hash from OpenSSL


AndreasFuchsSIT:
> Implement an `ameth` which implements `ASN1_PKEY_CTRL_DEFAULT_MD_NID`, reporting SHA256 as mandatory.

Split out from #15

AndreasFuchsSIT:
I have no clue on what an ameth is or where to register it.
This will need some investigation and (unless someone has good pointers) postponed until after 0.1 release.

AndreasFuchsSIT:
Notes from #15:

@AndreasFuchsSIT I'll let you ensure you have an ameth which implements ASN1_PKEY_CTRL_DEFAULT_MD_NID, reporting something like SHA1 or SHA256 as mandatory.

Arguably if you do that, it should be considered a bug in existing versions of OpenSSL that they then ask you to perform any other kind of signature. And in fact, that bug is fixed in my trees at
https://github.com/dwmw2/openssl/commits/pkey_md_1.1.1
https://github.com/dwmw2/openssl/commits/pkey_md_1.1.0
I'll let the PR for OpenSSL 1.1.2 settle first though, before I actually submit those for the stable branches (and I can do 1.0.2 too).

[...]

This is fixed in OpenSSL master now that openssl/openssl#7408 is merged.
PRs open for 1.0.2: openssl/openssl#7610 and 1.1.1: openssl/openssl#7609

This bug is all yours now :)

dwmw2:
I confess I don't really know. What you suggest looks reasonable.

AndreasFuchsSIT:
https://mta.openssl.org/pipermail/openssl-users/2018-November/009334.html pointed out that digest values are merely truncated if the curve uses fewer bits. Implement this alongside the ameth ctrl for maximum compatibility.

dwmw2:
The truncation seems to be working, although shouldn't it be using EC_GROUP_order_bits() instead of hard-coding it for just a couple of curves? That's what I've submitted to the other engine, in https://groups.io/g/openssl-tpm2-engine/topic/patch_truncate_hashes_for/32999767

I suggested that OpenSSL should perhaps do the truncation for itself: https://github.com/openssl/openssl/pull/9680

Note that if OpenSSL does that, then our trick of making up a hash type for the TPM to ignore based on the digest size is going to start failing. Perhaps we should just *always* say SHA256 regardless of the hash being used? Why does the TPM care anyway?

[...]

PR #72, opened 2019-01-25
ECDSA truncation and s_server test

AndreasFuchsSIT:
Implements truncation for ECDSA and adds an s_server test.
This test only works with OpenSSL 1.1.0 for some reason.

Addresses: #15 
Addresses: #31 

@dwmw2 Could you test this against your setup from #15 again ?

Does anybody know, why OpenSSL 1.0.2 attempts to compare private keys of the certificate ?
The error you get indicates such a behaviour...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -109,7 +109,8 @@ TESTS_SHELL = test/ecdsa.sh \
               test/failwrite.sh \
               test/rsasign_parent.sh \
               test/rsasign_persistent.sh \
-              test/rsasign_persistent_emptyauth.sh
+              test/rsasign_persistent_emptyauth.sh \
+              test/sserver.sh
 EXTRA_DIST += $(TESTS_SHELL)
 
 check_PROGRAMS = $(TESTS_UNIT)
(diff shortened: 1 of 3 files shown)


dwmw2:
Er, the test case you added *is* my test case from #15, isn't it?

The 1.0.2 failure rings a bell. Is it related to https://github.com/openssl/openssl/pull/1551 ?

AndreasFuchsSIT:
@dwmw2 Yep, I specifically meant inside your env as well. ;-)

I'll dig into the openssl issue you linked.

AndreasFuchsSIT:
@dwmw2 It sounds like it's similar and I have debuged it to the same point the you patched.
But running against your branch also did not work, no idea why...
```
Using default temp DH parameters
error setting private key
140201039440640:error:10071065:elliptic curve routines:EC_POINT_cmp:incompatible objects:ec_lib.c:1002:
140201039440640:error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch:x509_cmp.c:391:
```

AndreasFuchsSIT:
@dwmw2 It sounds like it's similar and I have debuged it to the same point the you patched.
But running against your branch also did not work...
```
Using default temp DH parameters
error setting private key
140201039440640:error:10071065:elliptic curve routines:EC_POINT_cmp:incompatible objects:ec_lib.c:1002:
140201039440640:error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch:x509_cmp.c:391:
```

Apparently, I'm entering a different case, where 0 was returned, not -2 from the previous functions.
Applying my patch to the case 0: did not help that much either, since I run into a later problem.

Maybe / Hopefully you have a better actual understanding, since I'm merely blindly stabbing at code without knowing what I'm doing here... ;-/

Issue #58, opened 2018-12-06
Fix StartUp Warning


AndreasFuchsSIT:
Remove the Esys_Startups from code and add it to test harness.

PR #73, opened 2019-01-25
Remove Esys_Startup from code

AndreasFuchsSIT:
Esys_Startup clutters the log with many errors.
A TPM should be assumed to be started up when used.
The test scripts will use tpm2-tools to start up the TPM in case
it wasn't.

Fixes: #58


AndreasFuchsSIT:
@PeterHuewe You can also leave reviews...

PeterHuewe:
I already approved the changes.

AndreasFuchsSIT:
Weird, github does not show and it's marked pending.
I'll override then...

Issue #70, opened 2019-01-23
Implement Symmetric Encryption AES


GaelGuegan:
Like the [tpm-tools]( https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_encryptdecrypt.1.md) provides the function **tpm2_encryptdecrypt** in order to make **symmetric encryption**, it would be nice to have the same here.

AndreasFuchsSIT:
Awesome...

PR #77, opened 2019-02-18
Add Symmetric Encryption

GaelGuegan:
### Description

Adding Symmetric Encryption : **tpm2-tss-engine-ciphers.c**. (Issue #70)

This is done by handling structure [EVP_CIPHER_meth_new](https://www.openssl.org/docs/manmaster/man3/EVP_CIPHER_meth_new.html).

### **How to use ?** 

1. Shell Command Line utility :
```shell
> Create a persistent sym key
> openssl enc -aes-256-cbc -e -engine tpm2tss -in data.txt -out enc_data -K 81000001 -iv 0123456789012345
> openssl enc -aes-256-cbc -d -engine tpm2tss -in enc_data -out dec_data.txt -K 81000001 -iv 0123456789012345
```

See the shell script : [ciphers.sh](https://github.com/GaelGuegan/tpm2-tss-engine/blob/master/test/ciphers.sh).

2. Openssl API in C code :
```c
strcpy(key , "81000001");
EVP_EncryptInit_ex(ctx, EVP_aes_256_cbc(), libtpm2tss, key, iv);
EVP_EncryptUpdate(ctx, pBufOut, &tmpLen, pBufIn, inLen);
EVP_EncryptFinal_ex(ctx, pBufOut + outLen, &tmpLen);
```

See official [wiki](https://wiki.openssl.org/index.php/EVP_Symmetric_Encryption_and_Decryption).

3. Create Key : 
```shell
tpm2tss-genkey -a aes -m cfb -s 128 sym_key
```

### Problems

For [now](https://github.com/tpm2-software/tpm2-tss-engine/pull/77#issuecomment-466006605), it is only working with **persistent handle**.  We **cannot** passed key file through the openssl app ([Doc](https://www.openssl.org/docs/man1.0.2/man1/openssl-enc.html)), only hex string.

Signed-off-by: Gael Guegan <gael.guegan@non.se.com>

### DIFF
--- a/src/tpm2-tss-engine-ciphers.c
+++ b/src/tpm2-tss-engine-ciphers.c
@@ -1,0 +1,385 @@
+/*******************************************************************************
+ * Copyright 2019, Schneider-Electric
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of tpm2-tss-engine nor the names of its contributors
+ * may be used to endorse or promote products derived from this software
+ * without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ ******************************************************************************/
+#define _DEFAULT_SOURCE // used for be32toh()
+#include <string.h>
+#include <endian.h>
+
+#include <openssl/engine.h>
+#include <openssl/evp.h>
+#include <tss2/tss2_esys.h>
+
+#include "tpm2-tss-engine.h"
+#include "tpm2-tss-engine-common.h"
+
+#define HANDLE_SIZE 8
+
+static int tpm2_cipher_nids[] = {
+    NID_aes_256_cfb128,
+    NID_aes_256_ofb128,
+    0
+};
+
+static int
+big_endian_array_to_native(const unsigned char *in, size_t size)
+{
+    uint32_t r;
+
+    memcpy(&r, in, size);
+    r = be32toh(r);
+
+    return r;
+}
+
+static int
+populate_tpm2data(const unsigned char *key, TPM2_DATA **tpm2Data)
+{
+    uint32_t keyHandle = 0;
+
+    /* Use an empty Key */
+    if (key == NULL) {
+        *tpm2Data = NULL;
+
+    /* Read persistent key, use for openssl app : openssl enc */
+    } else if ((key[0] == 0x81) && (key[HANDLE_SIZE/2 + 1] == 0)) {
+        keyHandle = big_endian_array_to_native(key, HANDLE_SIZE/2);
+        if (!tpm2tss_tpm2data_readtpm(keyHandle, tpm2Data))
+            return 0;
+
+    /* Use blob context */
+    } else {
+        return 0;
+    }
+
+    return 1;
+}
+
+static TPMI_ALG_SYM_MODE
+tpm2_get_cipher_mode(EVP_CIPHER_CTX *ctx, TPM2_DATA *tpm2Data)
+{
+    if (tpm2Data->pub.publicArea.parameters.symDetail.sym.mode.sym == TPM2_ALG_NULL) {
+        switch (EVP_CIPHER_CTX_mode(ctx)) {
+            case EVP_CIPH_CFB_MODE:
+                return TPM2_ALG_CFB;
+            case EVP_CIPH_OFB_MODE:
+                return TPM2_ALG_OFB;
+            case EVP_CIPH_CTR_MODE:
+                return TPM2_ALG_CTR;
+            case EVP_CIPH_ECB_MODE:
+                return TPM2_ALG_ECB;
+            case EVP_CIPH_CBC_MODE:
+                return TPM2_ALG_CBC;
+            default:
+                return TPM2_ALG_CBC;
+        }
+    } else {
+        return tpm2Data->pub.publicArea.parameters.symDetail.sym.mode.sym;
+    }
+}
+
+static int
+tpm2_cipher_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key, const unsigned char *iv, int enc)
+{
+    (void)(iv);
+    (void)(enc);
+    TPM2_DATA *tpm2Data = NULL;
+
+    DBG("Init Cipher Key ...\n");
+
+    /* Init Struct */
+    tpm2Data = OPENSSL_malloc(sizeof(TPM2_DATA));
+    if (tpm2Data == NULL) {
+        ERR(tpm2_cipher_init_key, ERR_R_MALLOC_FAILURE);
+        goto error;
+    }
+    memset(tpm2Data, 0, sizeof(TPM2_DATA));
+
+    /* Fill TPM2_DATA depending of KEY */
+    if (!populate_tpm2data(key, &tpm2Data)) {
+        ERR(tpm2_cipher_init_key, TPM2TSS_R_TPM2DATA_READ_FAILED);
+        goto error;
+    }
+
+    /* Set App Data */
+    EVP_CIPHER_CTX_set_app_data(ctx, tpm2Data);
+
+    DBG("algo : %#x", tpm2Data->pub.publicArea.parameters.symDetail.sym.algorithm);
+    DBG(" | mode : %#x", tpm2_get_cipher_mode(ctx, tpm2Data));
+    DBG(" | size : %d", tpm2Data->pub.publicArea.parameters.symDetail.sym.keyBits.sym);
+    DBG(" | enc  : %d", !EVP_CIPHER_CTX_encrypting(ctx));
+    DBG(" | iv   : %d\n", EVP_CIPHER_CTX_iv_length(ctx));
+
+    return 1;
+
+error :
+    if (tpm2Data)
+        OPENSSL_free(tpm2Data);
+    return 0;
+}
+
+static int
+tpm2_do_cipher(EVP_CIPHER_CTX *ctx, unsigned char *out, const unsigned char *in, size_t inl)
+{
+    TPM2_DATA *tpm2Data;
+    TSS2_RC ret;
+    ESYS_AUXCONTEXT eactx = (ESYS_AUXCONTEXT){ NULL, NULL};
+    ESYS_TR keyHandle = ESYS_TR_NONE;
+    TPM2B_MAX_BUFFER *out_data, in_data;
+    TPM2B_IV *iv_out, iv_in;
+    TPMI_ALG_SYM_MODE mode;
+    TPMI_YES_NO enc;
+
+    DBG("Ciphering ...\n");
+
+    /* Get App Data */
+    tpm2Data = EVP_CIPHER_CTX_get_app_data(ctx);
+    if (tpm2Data == NULL || tpm2Data->pub.size == 0 || inl == 0) {
+        return 0;
+    }
+
+    /* Init TPM key */
+    ret = init_tpm_key(&eactx, &keyHandle, tpm2Data);
+    ERRchktss(tpm2_do_cipher, ret, goto error);
+
+    /* Copy in_data : unsigned char* to TPM2B_MAX_BUFFER */
+    if (inl >= sizeof(in_data.buffer)) {
+        ERR(tpm2_do_cipher, TPM2TSS_R_GENERAL_FAILURE);
+        goto error;
+    }
+    memcpy(in_data.buffer, in, inl);
+    in_data.size = inl;
+
+    /* Get mode value */
+    mode = tpm2_get_cipher_mode(ctx, tpm2Data);
+
+    /* Get IV and Enc*/
+    iv_in.size = EVP_CIPHER_CTX_iv_length(ctx);
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+    memcpy(iv_in.buffer, ctx->iv, iv_in.size);
+    // Note: Openssl (encrypt:1) != TSS (encrypt:0)
+    enc = !(ctx->encrypt);
+#else
+    memcpy(iv_in.buffer, EVP_CIPHER_CTX_iv(ctx), iv_in.size);
+    enc = !EVP_CIPHER_CTX_encrypting(ctx);
+#endif
+
+    /* Trying to encrypt */
+    ret = Esys_EncryptDecrypt2( eactx.ectx,
+                                keyHandle,
+                                ESYS_TR_PASSWORD,
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                &in_data,
+                                enc,
+                                mode,
+                                &iv_in,
+                                &out_data,
+                                &iv_out );
+    if (ret == TPM2_RC_COMMAND_CODE) {
+        DBG("Esys_EncryptDecrypt2 : FAILED\n");
+        DBG("Failing back to Esys_EncryptDecrypt\n");
+        ret = Esys_EncryptDecrypt( eactx.ectx,
+                                   keyHandle,
+                                   ESYS_TR_PASSWORD,
+                                   ESYS_TR_NONE,
+                                   ESYS_TR_NONE,
+                                   enc,
+                                   mode,
+                                   &iv_in,
+                                   &in_data,
+                                   &out_data,
+                                   &iv_out );
+        if(ret == TPM2_RC_SUCCESS) {
+            DBG("Esys_EncryptDecrypt  : SUCCESS\n");
+        }
+        else {
+            DBG("Esys_EncryptDecrypt  : FAILED\n");
+        }
+    }
+    ERRchktss(tpm2_do_cipher, ret, goto error);
+
+    /* Copy out_data : TPM2B_MAX_BUFFER to unsigned char* */
+    memcpy(out, out_data->buffer, out_data->size);
+    out[out_data->size] = '\0';
+
+    /* Close TPM session */
+    if (keyHandle != ESYS_TR_NONE) {
+        if (tpm2Data->privatetype == KEY_TYPE_HANDLE) {
+            Esys_TR_Close(eactx.ectx, &keyHandle);
+        } else {
+            Esys_FlushContext(eactx.ectx, keyHandle);
+        }
+    }
+    esys_auxctx_free(&eactx);
+
+    return out_data->size;
+
+error :
+    /* Close TPM session */
+    if (keyHandle != ESYS_TR_NONE) {
+        if (tpm2Data->privatetype == KEY_TYPE_HANDLE) {
+            Esys_TR_Close(eactx.ectx, &keyHandle);
+        } else {
+            Esys_FlushContext(eactx.ectx, keyHandle);
+        }
+    }
+    esys_auxctx_free(&eactx);
+
+    if (tpm2Data)
+        OPENSSL_free(tpm2Data);
+    return 0;
+}
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+static EVP_CIPHER tpm2_aes_256_ofb =
+{
+    NID_aes_256_ofb128,                 // ID
+    TPM2_MAX_SYM_BLOCK_SIZE,            // Block size
+    TPM2_MAX_SYM_KEY_BYTES,             // Key length
+    TPM2_MAX_SYM_BLOCK_SIZE,            // IV length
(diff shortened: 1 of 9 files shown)


GaelGuegan:
Currently checking why CI failed ...

GaelGuegan:
After having trouble with sserver.sh, clang, openssl 1.0.2 ...

Finally, CI is passing. 

@AndreasFuchsSIT Can it be merged ?

AndreasFuchsSIT:
Glad you could make it work, I know the hassle... ;-)

I'll have a closer look over it after the 1.0 engine release. And queue it up for the 1.1 release...

From a first look, I think this still lacks tpm2tss-genkey support and loading of these, documentation in man-pages and bash-completion.

GaelGuegan:
I didn't find a way to use tpm2tss-genkey.  Because of the interface **openssl enc** which does not have options to read a key from a file. 

From the man pages of [openssl](https://www.openssl.org/docs/manmaster/man1/enc.html) :
> **-k password** :  the password to derive the key from. This is for compatibility with previous versions of OpenSSL. Superseded by the -pass argument.
**-kfile filename** :   read the password to derive the key from the first line of filename. This is for compatibility with previous versions of OpenSSL. Superseded by the -pass argument.
**-K key** :  The actual key to use: this must be represented as a string comprised only of hex digits. If only the key is specified, the IV must additionally specified using the -iv option. When both a key and a password are specified, the key given with the -K option will be used and the IV generated from the password will be taken. It does not make much sense to specify both key and password.

I was only able to use the option -K to pass a persistent handle of the TPM. Still looking for solutions ...

AndreasFuchsSIT:
Hmmmm..... If `-K` really requires the actual key, then the TPM handle does not suit this IMHO, because it's just the handle...

GaelGuegan:
I did that because we were doing the same with RSA. [See this test](https://github.com/tpm2-software/tpm2-tss-engine/blob/933d5d378a83b9b47599e452c703436bc5b4f39b/test/rsasign_persistent.sh#L37) : 
`echo "abc" | openssl pkeyutl -engine tpm2tss -keyform engine -inkey ${HANDLE} -sign -in ${DIR}/mydata.txt -out ${DIR}/mysig -passin stdin`

AndreasFuchsSIT:
I know, and I only justify it due to `-keyform engine`. Still no clean, but somewhat arguable.
Give me some time to think about this. Doc and Man would still be good though...

GaelGuegan:
I have made a working function tpm2tss_sym_genkey(), and write manual.

I am closing this temporarily, because I need to see with my company about copyright (I know nothing about this kind of stuff, sorry).

Link to my personal repo : https://github.com/GaelGuegan/tpm2-tss-engine

GaelGuegan:
CI passed if you click on details, don't know why it is blocked here

GaelGuegan:
I have made a local branch that fixed the **bug** when trying to encrypt string of **32 bytes or longer**.

However the CI my test is not passing with openssl version 1.0.2

src/tpm2tss-genkey.c
```diff
@@ -73,13 +74,15 @@ static const struct option long_options[] = {
...
+    {"symmode",  required_argument, 0, 'm'},
     {0,          0,                 0,  0 }
 };
 
 static struct opt {
     char *filename;
     TPMI_ALG_PUBLIC alg;
     TPMI_ECC_CURVE curve;
+    TPMI_ALG_SYM_MODE sym_mode;
     int exponent;
```
AndreasFuchsSIT: The help string, the longopts and this struct use different names, i.e. mode, symmode, sym_mode. Those should be unified.

src/tpm2tss-genkey.c
```diff
@@ -151,6 +158,20 @@ parse_opts(int argc, char **argv)
                 exit(1);
             }
             break;
+        case 'm':
+            if (strcasecmp(optarg, "cfb") == 0) {
+                opt.alg = TPM2_ALG_CFB;
```
AndreasFuchsSIT: Shouldn't this all be opt.sym_mode ?

src/tpm2tss-genkey.c
```diff
@@ -328,6 +378,10 @@ main(int argc, char **argv)
     case TPM2_ALG_ECDSA:
         tpm2Data = genkey_ecdsa();
         break;
+    case TPM2_ALG_AES:
+    case TPM2_ALG_CAMELLIA:
```
AndreasFuchsSIT: You sould add a line with `/* FALLTHROUGH */` here so static code analysis knows it is intentional.

GaelGuegan: OK

src/tpm2tss-genkey.c
GaelGuegan: aah yes sorry I have seen this mistake I have already fixed it in my other branch. Either I merge my branch in master (but CI not currently passing for openssl 1.0.2), either I correct it directly in master

GaelGuegan: also there are some other errors when managing sym mode

GaelGuegan:
@AndreasFuchsSIT Do you need something else ? more documentation ? more comments ?

AndreasFuchsSIT:
Looks like you are happily working on this.
Give me a heads up when you want me to review it.

GaelGuegan:
Okay does not matter I have made a patch anyway for myself, you can take your time

AndreasFuchsSIT:
Since 1.0.0 is out now, I'd like to start merging the backloged PRs.
How is it going on your side ? (When) Do you plan to rebase and squash your contributions to current master ?

GaelGuegan:
Do I need to rebase from my side, I think I have done it some time ago.

I didn't work on it, since last time.

GaelGuegan:
Ah ok I didn't see that there are conflicts, I will rebase.

GaelGuegan:
@AndreasFuchsSIT Does it not pose a problem that I was not able to make it works with "tpm2tss-genkey" file ? I can create the file but don't how to get it through openssl. (see https://github.com/tpm2-software/tpm2-tss-engine/pull/77#issuecomment-466006605)

AndreasFuchsSIT:
@GaelGuegan you merged but you did not rebase. Could you rebase and squash your contribution such that they are a single commit on top of current master ? That would make things much easier.

Good point about the "cannot read key from file". That is a problem. I guess that actually does still require a solution to be found. :-(

GaelGuegan:
I have rebased and squashed as you asked, and CI is passing. 

I did not do it on purpose but there is twice my name in the commit, because of my wrong git configuration sorry.

GaelGuegan:
@AndreasFuchsSIT I have rebased and squashed, is that ok for you ?

AndreasFuchsSIT:
Yes, that's great...
Now let's figure out how to get the TPM keyids in there...
Maybe we could ask openssl upstream developers if they have something in mind...

GaelGuegan:
Currently, I don't think that the **openssl app** can pass file. Maybe that in the meantime that openssl developpers find a solution, if there is one, shouldn't we merge this PR ?

However, I think it would be possible to use our key file through **openssl API** : EVP_Encrypt(...).
see [wiki](https://www.openssl.org/docs/manmaster/man3/EVP_EncryptInit.html) and [man](https://wiki.openssl.org/index.php/EVP_Symmetric_Encryption_and_Decryption).

GaelGuegan:
Maybe, one solution would be to get our filename, open and read it in our engine. But the filename need to be a hexadecimal string because of that [function](https://github.com/openssl/openssl/blob/master/apps/enc.c#L647).

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,527 @@
...
+
+    /* Read persistent key, use for openssl app : openssl enc */
+    } else if ((key[0] == 0x81) && (key[HANDLE_SIZE/2 + 1] == 0)) {
+        keyHandle = convert_array_hex_to_int(key, HANDLE_SIZE/2);
+        if (!tpm2tss_tpm2data_readtpm(keyHandle, tpm2Data))
+            return 0;
+
+    /* Use blob context */
+    } else {
+        //if (!tpm2tss_tpm2data_read((char *)key, tpm2Data))
```
AndreasFuchsSIT: If we want to merge the current version, this should probably be removed.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,527 @@
...
+    for(size_t i = 0; i < size; i++)
+    {
+        integer |= in[i] << 8*(size - i - 1);
+    }
+
+    return integer;
+}
+
+int
+tpm2tss_sym_genkey(EVP_CIPHER_CTX *cipher, TPMI_ALG_PUBLIC algo,
```
AndreasFuchsSIT: If we only support persistent keys for now, then this function should be removed

src/tpm2tss-genkey.c
```diff
@@ -50,7 +50,8 @@ char *help =
     "Arguments:\n"
     "    <filename>      storage for the encrypted private key\n"
     "Options:\n"
-    "    -a, --alg       public key algorithm (rsa, ecdsa) (default: rsa)\n"
+    "    -a, --alg       public key algorithm (rsa, ecdsa, aes) (default: rsa)\n"
```
AndreasFuchsSIT: If we only support persistent keys for now, then all changes to this file should be removed.

test/ciphers.sh
```diff
@@ -0,0 +1,60 @@
...
+    openssl enc -aes-256-cbc -e -engine tpm2tss -in ${DIR}/data1.txt -out ${DIR}/enc_data1 -K ${KEY} -iv ${IV}
+    openssl enc -aes-256-cbc -d -engine tpm2tss -in ${DIR}/enc_data1 -out ${DIR}/dec_data1 -K ${KEY} -iv ${IV}
+    diff ${DIR}/data1.txt ${DIR}/dec_data1
+fi
+
+# Release persistent HANDLE
+tpm2_evictcontrol -a o -c ${HANDLE}
+
+# Test Gen Key
+tpm2tss-genkey -a aes -m cfb -s 128 ${DIR}/mykey
```
AndreasFuchsSIT: This should be removed, since there is no way for us to use the generated key afterwards.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,527 @@
...
+populate_tpm2data(const unsigned char *key, TPM2_DATA **tpm2Data)
+{
+    uint32_t keyHandle = 0;
+
+    /* Use an empty Key */
+    if (key == NULL) {
+        tpm2Data = NULL;
+
+    /* Read persistent key, use for openssl API : EVP_EncryptInit_ex() */
+    } else if (strncmp((char *)key, "0x81", 4) == 0) {
```
AndreasFuchsSIT: Since this is never used, we should just remove it

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,527 @@
...
+static int tpm2_cipher_nids[] = {
+    NID_aes_192_cbc,
+    NID_aes_256_cfb128,
+    NID_aes_256_cbc,
+    NID_aes_256_ofb128,
+    0
+};
+
+static int
+convert_array_hex_to_int(const unsigned char *in, size_t size)
```
AndreasFuchsSIT: This naming is misleading, since it's not a hex array. Rather this is a `big-endian-array-to-native` conversion. My recommended code to also prevent page-align problems would be:
```
if (size != 4) fail;
uint32 r;
memcpy(&r, in, 4);
r = be32toh(r);
```

man/tpm2tss_sym_genkey.3.md
```diff
@@ -0,0 +1,38 @@
+% tpm2tss-tpm2data_write(1) tpm2-tss-engine | Library calls
```
AndreasFuchsSIT: This file could probably go if we support persistent only

man/tpm2tss-genkey.1.md
```diff
@@ -24,14 +24,17 @@ key information. This file can then be loaded with OpenSSL using
 # OPTIONS
```
AndreasFuchsSIT: All these changes can go, if we support only persistent

src/tpm2-tss-engine-ciphers.c
GaelGuegan: Okay I needed to add the flag : AX_ADD_PREPROC_FLAG([-D_DEFAULT_SOURCE]) in order to use the function be32toh() with <endian.h>

src/tpm2-tss-engine-ciphers.c
GaelGuegan: Previous code : 
```c
    /* Use blob context */
    } else {
        //if (!tpm2tss_tpm2data_read((char *)key, tpm2Data))
        //    return 0;
    }
return 1;
```
Current code :
```c
    /* Use blob context */
    } else {
        return 0;
    }
return 1;
```
Because of that, `tpm2_cipher_init_key()` is failing when doing the `test/sserver.sh` with openssl 1.1.0.

GaelGuegan:
CFB and OFB are now the two symmetric modes available. CBC was apparently causing troubles with openssl sserver & sclient. 

However if a mode is already defined within the TPM persistent key, it will use it.

GaelGuegan:
:wink:

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+        return 0;
+    }
+
+    /* Init TPM key */
+    ret = init_tpm_key(&eactx, &keyHandle, tpm2Data);
+    ERRchktss(tpm2_do_cipher, ret, goto error);
+
+    /* Copy in_data : unsigned char* to TPM2B_MAX_BUFFER */
+    in_data = OPENSSL_malloc(sizeof(TPM2B_MAX_BUFFER));
+    if (tpm2Data == NULL) {
```
AndreasFuchsSIT: you mean `in_data` ?

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+}
+
+static int
+tpm2_do_cipher(EVP_CIPHER_CTX *ctx, unsigned char *out, const unsigned char *in, size_t inl)
+{
+    TPM2_DATA *tpm2Data;
+    TSS2_RC ret;
+    ESYS_AUXCONTEXT eactx = (ESYS_AUXCONTEXT){ NULL, NULL};
+    ESYS_TR keyHandle = ESYS_TR_NONE;
+    TPM2B_MAX_BUFFER *out_data, *in_data;
```
AndreasFuchsSIT: in_data is never freed.
Also you don't have to malloc it, you could just declare `TPM2_MAX_BUFFER in_data;` on stack instead of a pointer.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+    ret = init_tpm_key(&eactx, &keyHandle, tpm2Data);
+    ERRchktss(tpm2_do_cipher, ret, goto error);
+
+    /* Copy in_data : unsigned char* to TPM2B_MAX_BUFFER */
+    in_data = OPENSSL_malloc(sizeof(TPM2B_MAX_BUFFER));
+    if (tpm2Data == NULL) {
+        ERR(tpm2_do_cipher, ERR_R_MALLOC_FAILURE);
+        goto error;
+    }
+    memcpy(in_data->buffer, in, inl);
```
AndreasFuchsSIT: You should check that `inl < sizeof(in_data->buffer)` before the memcpy...

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                in_data,
+                                enc,
+                                mode,
+                                &iv_in,
+                                &out_data,
+                                &iv_out );
+    if (ret == TPM2_RC_COMMAND_CODE) {
+        DBG("Esys_EncryptDecrypt2 : FAILED\n");
```
AndreasFuchsSIT: I'd add "Falling back to Esys_EncryptDecrypt" or something like that.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+    }
+    esys_auxctx_free(&eactx);
+
+    if (tpm2Data)
+        OPENSSL_free(tpm2Data);
+    return 0;
+}
+
+static int
+tpm2_cipher_cleanup(EVP_CIPHER_CTX *ctx)
```
AndreasFuchsSIT: Since this is basically a no-op, can you just not declare this and pass NULL in the struct ?

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+{
+    (void)(ctx);
+    DBG("Cleaning up ...\n");
+    return 1;
+}
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+static EVP_CIPHER tpm2_aes_256_ofb =
```
AndreasFuchsSIT: Sorry for being so ignorant, but could you explain the 256 and ofb128 to me ?
I don't know what they mean.
Does that mean that aes 128 is not supported ?

I guess a comment in code would also be good...

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+{
+    NID_aes_256_ofb128,                 // ID
+    TPM2_MAX_SYM_BLOCK_SIZE,            // Block size
+    TPM2_MAX_SYM_KEY_BYTES,             // Key length
+    TPM2_MAX_SYM_BLOCK_SIZE,            // IV length
+    EVP_CIPH_OFB_MODE,                  // Flags
+    tpm2_cipher_init_key,               // Init key
+    tpm2_do_cipher,                     // Encrypt/Decrypt
+    tpm2_cipher_cleanup,                // Cleanup
+    sizeof(TPM2_DATA),                  // Context size
```
AndreasFuchsSIT: I don't think this is correct. The man-page says `EVP_CIPHER_meth_set_impl_ctx_size() sets the size of the EVP_CIPHER's implementation context so that it can be automatically allocated.` but we are malloc'ing it ourselves. But I also don't know what this is really for and how we should use it.

test/ciphers.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -eufx
+
+export OPENSSL_ENGINES=${PWD}/.libs
```
AndreasFuchsSIT: There were some simplification merged for the test scripts in a recent PR.
Please look at the current examples https://github.com/tpm2-software/tpm2-tss-engine/blob/master/test/rsasign.sh

AndreasFuchsSIT:
> 
> 
> CFB and OFB are now the two symmetric modes available. CBC was apparently causing troubles with openssl sserver & sclient.
> 
> However if a mode is already defined within the TPM persistent key, it will use it.

I guess that had to do with the comment I posted here: https://github.com/tpm2-software/tpm2-tss-engine/pull/77#discussion-diff-281226821R163

Please see that we enable CBC mode for the TPM and still have sserver running. It's basically the counter test that the engine does not interfere with any other operations; especially if the engine shall only be used for asymmetric operations.
Also I don't know about the switch-case statement in the selector function. Maybe the default there also needs to be a standard engine.

AndreasFuchsSIT:
As a final comment: Since only very few TPMs support symmetric operations, I'd like to wrap support for it in a configure parameter `--enable-symmetric`; just to minimize interference possibilities with the asymmetric operations.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+populate_tpm2data(const unsigned char *key, TPM2_DATA **tpm2Data)
+{
+    uint32_t keyHandle = 0;
+
+    /* Use an empty Key */
+    if (key == NULL) {
+        *tpm2Data = NULL;
+
+    /* Read persistent key, use for openssl app : openssl enc */
+    } else if ((key[0] == 0x81) && (key[HANDLE_SIZE/2 + 1] == 0)) {
```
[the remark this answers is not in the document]
GaelGuegan: That not possible here. `strnlen(key, HANDLE_SIZE*2+1)` will always return 1 because the second byte of data is `0x00` in a key handle, for example `0x81000001`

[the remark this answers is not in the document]
AndreasFuchsSIT: Oh sorry, of course you are right.

src/tpm2-tss-engine-ciphers.c
GaelGuegan: To be honest, I don't fully understand this stuff neither.

I based my code by taking the [gost engine](https://github.com/gost-engine/engine) as example. In there code, we can see that there are doing a `EVP_CIPHER_meth_set_impl_ctx_size(_hidden_magma_cbc, sizeof(struct ossl_gost_cipher_ctx))` [here](https://github.com/gost-engine/engine/blob/master/gost_crypt.c#L248) and then `struct ossl_gost_cipher_ctx *c = EVP_CIPHER_CTX_get_cipher_data(ctx);` [here](https://github.com/gost-engine/engine/blob/master/gost_crypt.c#L461).

So maybe I should put a 0 value here, I tried and it does not seem to change anything. I can change it if you want ?

src/tpm2-tss-engine-ciphers.c
GaelGuegan: So this is about the **selector function** and **NIDS**, I guess. So will try to explain my understanding of it.

So at first, you can try to type a `openssl enc -ciphers` on your machine. You will see there the different openssl symmetric encryption available. We have to define which algos of this list , our engine will implement. Currently, I have defined these two : 
```c
static int tpm2_cipher_nids[] = {
    NID_aes_256_cfb128,
    NID_aes_256_ofb128,
    0
};
```
Which are defined by AES algorithm, 256 bits key size and cfb/ofb encryption mode.
When someone call `openssl enc -aes-256-cfb -e -engine tpm2tss ...`, our engine implementation will be used.

Moreover, IDs are defined [here](https://github.com/openssl/openssl/blob/master/include/openssl/obj_mac.h#L2735) in the openssl repo. 

As i said before I took the gost engine as example, we can see [here](https://github.com/gost-engine/engine/blob/master/gost_eng.c#L345) the implementation of its cipher selector. Also another really simple cipher engine exits [here](https://github.com/bigbrett/wsaesengine/blob/master/src/wsaesengine.c#L193).

GaelGuegan: If you have other question, don't hesistate. I will try to be the clearest possible and explain my understanding.

src/tpm2-tss-engine-ciphers.c
```diff
@@ -0,0 +1,393 @@
...
+    TPM2B_IV *iv_out, iv_in;
+    TPMI_ALG_SYM_MODE mode;
+    TPMI_YES_NO enc;
+
+    DBG("Ciphering ...\n");
+
+    /* Get App Data */
+    tpm2Data = EVP_CIPHER_CTX_get_app_data(ctx);
+    if (tpm2Data == NULL || tpm2Data->pub.size == 0 || inl == 0) {
+        return 0;
```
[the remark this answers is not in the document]
GaelGuegan: mmmh don't know of to do that here.

And if tpm2Data is NULL, tpm2_cipher_init_key() function should has failed before.

GaelGuegan:
I have done some changes that you requested. However can you relaunch my CI build ? a weird error appeared, CI was not able to clone openssl.

AndreasFuchsSIT:
You can't but I did

AndreasFuchsSIT:
Since https://github.com/tpm2-software/tpm2-tools/pull/1476 I'll retrigger the CI...

GaelGuegan:
don't understand why CI is not passing here, because it passed in my personnal branch [here](https://travis-ci.org/GaelGuegan/tpm2-tss-engine)

AndreasFuchsSIT:
Looks very much stalled. So closing

Issue #79, opened 2019-02-18
Tests rsasign_persistent and rsasign_persistent_emptyauth hang indefinitely with tpm2-tss 2.2.0


diabonas:
Building with tpm2-tss 2.2.0 causes the integration tests [`test/rsasign_persistent.sh`](https://github.com/tpm2-software/tpm2-tss-engine/blob/master/test/rsasign_persistent.sh) and [`test/rsasign_persistent_emptyauth.sh`](https://github.com/tpm2-software/tpm2-tss-engine/blob/master/test/rsasign_persistent_emptyauth.sh) to hang indefinitely while they are trying to execute
```
openssl pkeyutl -engine tpm2tss -keyform engine -inkey ${HANDLE} -sign -in ${DIR}/mydata.txt -out ${DIR}/mysig -passin stdin
```
The problem seems to be related to the [tpm2-tss compatibility code in `tpm2-tss-engine-common.c`](https://github.com/tpm2-software/tpm2-tss-engine/blob/9b0c7d1225f194a6098ceabcc9b0392abf8e1deb/src/tpm2-tss-engine-common.c#L243): if the `#ifdef TSS22`...`#endif` part is removed, everything works as expected.

diabonas:
This seems to be related to the [switch of the crypto backend to OpenSSL in tpm2-tss 2.2.0](https://github.com/tpm2-software/tpm2-tss/commit/a5ff71abeda9a5aea269812de32afb79ef3fcb4d): if I compile tpm2-tss 2.2.0 with `./configure --with-crypto=gcrypt`, everything works as expected. I guess this makes this a tpm2-tss bug rather than a tpm2-tss-engine issue. However I am currently unable to provide an example that reproduces the bug in tpm2-tss without the use of tpm2-tss-engine.

AndreasFuchsSIT:
Could you post your openssl version ?

And could you run it with TSS2_LOG=esys+trace so we see on which function it hangs ?

diabonas:
I am running OpenSSL 1.1.1a on Arch Linux. The call seems to get stuck at [`Esys_StartAuthSession`](https://github.com/tpm2-software/tpm2-tss-engine/blob/9b0c7d1225f194a6098ceabcc9b0392abf8e1deb/src/tpm2-tss-engine-common.c#L256):
```
[...]

PR #82, opened 2019-02-19
Fix parallel connections to TPM.

AndreasFuchsSIT:
Esys internal us of the TPM conflicted with our lease of the TPM, when using sessions under tpm2tss >=2.2.0. Fixing this by overriding the random engine to not be ourselves.

Fixes: #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ dist: trusty
 
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


AndreasFuchsSIT:
@diabonas Could I get your review and testing on this ?
Highly appreciated !

PR #97, opened 2019-03-21
configure.ac: ensure that for the integration tests tpm2_startup is installed

dilyanpalauzov:


AndreasFuchsSIT:
Actually, tpm2_startup should be moved to the travis.yml scrip.

But tpm2-tools need to be check in general, since there's more tools used here than just tpm2_startup, so you are right in that regard...

AndreasFuchsSIT:
Do you want to add the tools tpm2_createprimary, tpm2_flushcontext, tpm2_evictcontrol, tpm2_create, tpm2_load, tpm2_readpublic as well or shall I ?

dilyanpalauzov:
Ensuring that tpm2_startup is installed in fact makes sure, that tpm2-tools is installed.  There is no way to install tpm2_startup without tpm2_createprimary, so checking for available tpm2_createprimary, in addition to tpm2_startup, is overengineering.

PR #110, opened 2019-04-24
Some more updates to tpm2-tools

AndreasFuchsSIT:
I've had a fix and some more stuff ready, but @diabonas was quicker.
Here are my remainders:
- Using long parameters for the tools, hoping to be more stable
- Add some text to INSTALL.md about tools and integration testing.


AndreasFuchsSIT:
@diabonas Can I consider the ThumbsUp as positive review ?
In all honesty, if you could post (positive) reviews as comments, I'd consider them sufficient for merging at this point!

PR #114, opened 2019-05-09
Add ccache to speed up travis runs.

DomiNic5787:
see [https://github.com/tpm2-software/tpm2-tss/pull/1403](https://github.com/tpm2-software/tpm2-tss/pull/1403)

Signed-off-by: Dominic Grauvogl <dominicmanuel.grauvogl@infineon.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ compiler:
 
 dist: xenial
 
+cache: ccache
+
 env:
   matrix:
   - OPENSSL_BRANCH=OpenSSL_1_0_2-stable TPM2TSS_BRANCH=2.2.x


AndreasFuchsSIT:
Ok, I guess tpm2-tools changed again, because the variable for the inkey is empty in the test...
I will look into this soon...

AndreasFuchsSIT:
Since https://github.com/tpm2-software/tpm2-tools/pull/1476 I'll retrigger the CI...

AndreasFuchsSIT:
@DomiNic5787 Maybe you know more here.
Apparendly, our ccache is broken now, because it carries a corrupted libcrypto.so, see https://travis-ci.org/tpm2-software/tpm2-tss-engine/jobs/540671744

At least that's my best guess following https://stackoverflow.com/questions/5713894/file-not-recognized-file-truncated-gcc-error

Do you have any idea, how we can clear the ccache of travis ?
Because I don't know how to do that...

DomiNic5787:
@AndreasFuchsSIT 
Description to delete the cache in travis: https://docs.travis-ci.com/user/caching/#clearing-caches

AndreasFuchsSIT:
Looks very good.
Danke !

DomiNic5787:
@AndreasFuchsSIT 
A feature like https://github.com/travis-ci/travis-ci/issues/4472 could avoid this issue.
Unfortunately it was never implemented.

AndreasFuchsSIT:
I took the liberty to mention it there. Maybe it triggers someone... ;-)
Thanks for the help!

diabonas:
I think the problem is not even a failed build, I can consistently reproduce this with [tpm2-totp](https://github.com/tpm2-software/tpm2-totp) even if all builds are successful: the [first build](https://travis-ci.org/diabonas/tpm2-totp/builds/541266470) with empty cache succeeds, [subsequent builds](https://travis-ci.org/diabonas/tpm2-totp/builds/541271955) with initialised cache fail with
```
../libcrypto.so: file not recognized: File truncated
```
I suspect this is a problem with the OpenSSL Makefile not properly supporting parallel jobs, since [using `make -j1` for OpenSSL](https://travis-ci.org/diabonas/tpm2-totp/builds/541276470) succeeds even with the same previously failing cache. The speedup gained from this is about two minutes total time for tpm2-totp (i.e. the GCC job is about two minutes faster, Clang doesn't seem to use ccache at all).

As an alternative, I have experimented with [caching the complete dependency folders](https://github.com/diabonas/tpm2-totp/commit/90359fe089a65855c23c2b145316d4c190953728), relying on `make` to rebuild only the necessary files based on timestamps. This [works](https://travis-ci.org/diabonas/tpm2-totp/builds/541293592) and gives a nice speedup of about six minutes total time (i.e. both GCC and Clang are about three minutes faster), but I'm still testing and thinking about the best implementation (e.g. how to clear the cache if you're switching to a different branch for tpm2-tools etc.).

PR #115, opened 2019-05-20
build: link against tss2-mu

diabonas:
Currently the linking of `libtpm2tss` relies on the fact that the upstream tpm2-tss pkg-config file explicitly requires tss2-mu. Since this is wrong and will be changed upstream, but we need access to the marshalling functions, explicitly link against tss2-mu. See https://github.com/tpm2-software/tpm2-tss/pull/1417 for a more detailed discussion.


diabonas:
A fix for the CI failure is in #116.

AndreasFuchsSIT:
Merged. Wanna rebase ?

PR #117, opened 2019-05-30
build: add --disable-defaultflags

alonbl:
Align with tpm2-tss to allow downstream to provide flags.


alonbl:
@diabonas this is similar issue.

PR #118, opened 2019-05-31
added VxWorks support

coolfengyu:
Signed-off-by: yfeng <yu.feng@windriver.com>

PR #119, opened 2019-05-31
 test: use tpm2-tools 3.X

diabonas:
Since https://github.com/tpm2-software/tpm2-tools/commit/893e43271024e9c2d942c8dc9ba1657a667bb0e2 tpm2-tools explicitly requires tpm2-tss >=2.3.0, so we switch to the stable version of tpm2-tools instead. This has been attempted previously in #61, but failed due to the missing `tpm2_flushcontext` tool. Since that tool has been [backported to tpm2-tools 3.X](https://github.com/tpm2-software/tpm2-tools/pull/1471), we can now use it for the CI. Therefore this is basically a rewrite of #61 for the long option names introduced in https://github.com/tpm2-software/tpm2-tss-engine/pull/110.

Note that tpm2-tools 3.2 that will [include tpm2_flushcontext](https://lists.01.org/pipermail/tpm2/2019-May/001195.html) is not yet released, so users can't run the integration tests with the tpm2-tools package provided by their distribution. Until that happens, they will manually have to build the [tpm2-tools 3.X branch](https://github.com/tpm2-software/tpm2-tools/tree/3.X) instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ dist: xenial
 
 env:
   matrix:
-  - OPENSSL_BRANCH=OpenSSL_1_0_2-stable TPM2TSS_BRANCH=2.2.x
-  - OPENSSL_BRANCH=OpenSSL_1_1_0-stable TPM2TSS_BRANCH=2.2.x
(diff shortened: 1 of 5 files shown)


diabonas:
This fixes the CI failures observed by #117 and #118.

PR #123, opened 2019-06-04
added NO_DL build configuration

coolfengyu:
added NO_DL build configuration. If defined NO_DL, the engine will not support using the tcti shared library. Some system such as VxWorks builds the tcti as static library and links with OS image together. In this situation, shared library is not supported and should be removed from engine.

### DIFF
--- a/src/tpm2-tss-engine-tcti.c
+++ b/src/tpm2-tss-engine-tcti.c
@@ -73,7 +73,7 @@ tcti_expand_dlname(const char *shortname)
 static TSS2_RC
 tcti_dlopen(const char *dl_path, dl_handle_t *dlhandle_p)
 {
-    TSS2_RC r = TSS2_RC_SUCCESS;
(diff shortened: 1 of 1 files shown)


src/tpm2-tss-engine-tcti.c
```diff
@@ -68,7 +73,7 @@ tcti_expand_dlname(const char *shortname)
 static TSS2_RC
 tcti_dlopen(const char *dl_path, dl_handle_t *dlhandle_p)
 {
-    TSS2_RC r;
+    TSS2_RC r = TSS2_RC_SUCCESS;
```
AndreasFuchsSIT: I don't like unnecessary initialization of variables. I prefer to have my compiler/analyzer tell me that I missed a case, which is not the case here. So please revert this line.

coolfengyu:
The latset code restored tcti_set_opts ().

PR #125, opened 2019-06-05
added NO_DL build configuration.

coolfengyu:
If defined NO_DL, the engine will not support using the tcti shared library.
Some system such as VxWorks builds the tcti as static library and links with OS image together.
In this situation, shared library is not supported and should be removed from engine.

### DIFF
--- a/src/tpm2-tss-engine-tcti.c
+++ b/src/tpm2-tss-engine-tcti.c
@@ -1,6 +1,8 @@
 //**********************************************************************;
 // Copyright (c) 2018, General Electric Company.
 // All rights reserved.
+// Copyright (c) 2019, Wind River Systems.
+// All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -28,7 +30,9 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#ifndef NO_DL
 #include <dlfcn.h>
+#endif /* NO_DL */
 
 #include <tss2/tss2_tcti.h>
 #include <openssl/err.h>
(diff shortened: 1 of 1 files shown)


coolfengyu:

AndreasFuchsSIT:
IMO, this patch is too narrow.
Instead the whole engine-control should be deactivated instead and I guess the complete -tcti.c file excluded from build.
I mean, what purpose does the engine-crtl serve, if there is no backing code ?

coolfengyu:
Good suggestion. How about create a new file tpm2-tss-engine-static-tcti.c? It will use the tcti static library.

AndreasFuchsSIT:
The uestion is, whether we want to statically set a TCTI for the engine here, or whether we just want to rely on ESYS to set a static TCTI correctly ?

coolfengyu:
The engine should rely on ESYS to set a static TCTI by default. Should we consider the different TCTI configuration or not? If yes, the dynamic and static libraries should be considered also.

AndreasFuchsSIT:
Since you're the only user of NO_DL at this moment: You tell me, if you need a TCTI-ctrl for static.

In either case, the current TCTI-ctrl needs should be disabled alltogether for no-dynamic builds and I guess the complete file tpm2-tss-engine-tcti.c should be excluded from the build.

It should take a configure-flag as well, I guess.

coolfengyu:
The ESYS default configuration is enough for me. It can work with VxWorks. 
tpm2-tss-engine-tcti.c includes some public APIs, such as tcti_clear_opts() and tcti_set_opts(). Currently, it can't be removed for no-dynamic builds.
In addition, some functions in other .c files need to be modified (i.e esys_auxctx_init(), esys_auxctx_free().)
It seems the ESYS initializing and finalizing processes need to be redesigned.

AndreasFuchsSIT:
Hmpf... sounds like it... :-(

AndreasFuchsSIT:
We may run into a redesign of this anyways with the tss2-tctildr.so library appearing on tpm2-tss.
Thus, I guess we can postpone this until that library is available upstream.

AndreasFuchsSIT:
Seems stalled. So closing.

tctildr is a thing now though, so might make this whole PR much simpler.

PR #126, opened 2019-06-05
This patch adds engine load routine for OpenSSL.

coolfengyu:
If OpenSSL doesn't support load the engines dynamicly,
we need to provide a special load routine so that the applications could call this routine for loading the engine by themself.

### DIFF
--- a/include/tpm2-tss-engine.h
+++ b/include/tpm2-tss-engine.h
@@ -1,6 +1,8 @@
 /*******************************************************************************
  * Copyright 2017-2018, Fraunhofer SIT sponsored by Infineon Technologies AG
  * All rights reserved.
+ * Copyright (c) 2019, Wind River Systems.
+ * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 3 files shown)


coolfengyu:
If OpenSSL is static, this engine will be static.
It has been tested on VxWorks.
For other platforms, the test case should call ENGINE_load_tpm2tss before using this engine and check whether the engine works correctly.

coolfengyu:

AndreasFuchsSIT:
As mentioned, could you add a case to .travis.yml or give me a sequence on how to test this static linking ?

coolfengyu:
When configuring openssl, the test case needs to add "no-dynamic-engine".
For example, ./config --prefix=/usr/local --openssldir=/usr/local/openssl shared no-dynamic-engine

Then, ENGINE_load_tpm2tss() needs to be added to ENGINE_load_builtin_engines() in openssl.

AndreasFuchsSIT:
Any idea why travis is failing ?

coolfengyu:
This testing needs to create patch for OpenSSL. I try to investigate whether vxworks supporint dynamic-engine. I will close this merge request.

PR #127, opened 2019-06-05
Enhanced init_tpm_parent(). It could use RSA algorithm to create primary key.

coolfengyu:
It could use RSA algorithm to create primary key.
Some old TPM 2.0 devices don't support ECC, such as Infineon SLB9665TT2.0 TPM version 2.0.
This init_tpm_parent() will check the TPM device whether support ECC. If it supports, it uses ECC. Otherwise, it will use RSA.

### DIFF
--- a/src/tpm2-tss-engine-common.c
+++ b/src/tpm2-tss-engine-common.c
@@ -1,6 +1,8 @@
 /*******************************************************************************
  * Copyright 2017-2018, Fraunhofer SIT sponsored by Infineon Technologies AG
  * All rights reserved.
+ * Copyright (c) 2019, Wind River Systems.
+ * All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


PeterHuewe:
One minor correction to your commit message.
There is no _version 2.0_ of SLB 9665 TPM2.0. The FW Version without ECC is 5.00
`TPM_PT_FIRMWARE_VERSION_1: 0x00050000`
Newer versions of course have ECC - but there might be other TPMs from other vendors who also don't support ECC, so the commit itself makes sense.

coolfengyu:

AndreasFuchsSIT:
Please change the commit message according to PeterHuewe's comment.
Then I will merge.

coolfengyu:
OK, the commit message has been updated.

PR #134, opened 2019-07-10
Add support for using a parent key password

khwaaj:
This PR adds support for providing a parent key password to tpm2tss-genkey and extends the engine to support setting it using an ENGINE_ctrl() command (TPM2TSS_SET_PARENTAUTH or "SET_PARENTAUTH").

PR #135, opened 2019-07-18
Support setting binary passwords

khwaaj:
Add ability to use binary passwords by passing them as a prefixed hex-string, much like the tpm2-tools package already supports. This includes both the tpm2tss-genkey tool and the SET_OWNERAUTH/SET_PARENTAUTH ctrls, where the ctrls also support directly using data not terminated by a null by passing the data length in the numeric argument of the ctrl.

#### Compatibility Analysis
There is a small chance the changes can break existing use-cases under some very specific circumstances:

1. For someone using a password starting with "hex:" or "str:". This applies to both the command-line options and the ENGINE ctrls. The alternative here would be to have separate command-line options and ctrls, but that would set the solution apart from what is currently done by tpm2-tools.

2. For someone using the ENGINE ctrls and passing a non-zero, non-strlen(), value as the numeric argument, even though it is currently unused. Since passing 0/NULL for unused arguments is established practice I don't see the risk of this being very high. The alternative would be to introduce separate ctrls with the new numeric argument behaviour.

All in all, my opinion is that the risks mentioned above are minimal and that keeping the interfaces similar to tpm2-tools is a higher goal in this instance, especially for (1). As mentioned in PR #134, writing a tpm2tss-engine.7 manpage is probably a good idea now that the ctrls are getting more complicated, I'll see if I can find the time to put one together and add to this PR if it is accepted. Let me know what you think!

### DIFF
--- a/src/tpm2-tss-engine-common.c
+++ b/src/tpm2-tss-engine-common.c
@@ -33,6 +33,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <ctype.h>
 
 #include <openssl/engine.h>
 #include <openssl/pem.h>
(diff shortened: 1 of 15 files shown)


AndreasFuchsSIT:
Sorry for the long delay. I appreciate you working on this.

Quick uestion: What happens if neither str: nor hex: are provided. From the manpage updates it seems that this would be illegal. From the testsuit it seems to default to str. Could you clarify this for me ?

khwaaj:
Sure, the key is in the phrase "Unless otherwise specified passwords are assumed to be in string form", though perhaps it could be clearer. Basically the default is to interpret as a regular string password. If there is a "hex:" prefix the following characters are decoded as hexadecimal. The "str:" prefix is only there to make things work in the unlikely case that a password starting with "hex:" is used (same as in the tmp2-tools package).

AndreasFuchsSIT:
Maybe you could put this non-prefix explanation into the bullet-point list as well, so others don't stumble over it as I did.
Also some more words in the function description and commit message would be nice.

khwaaj:
Sorry for the delay but I finally got around to making some improvements to the things you suggested. Have a  look and tell me what you think!

AndreasFuchsSIT:
Sorry for screwing you over, but after merging #142 this PR won't work anymore due to the change of tpm2-tools.
Could you rebase against master and make the necessary fixes ?

src/tpm2-tss-engine-common.c
```diff
@@ -55,6 +56,84 @@ IMPLEMENT_ASN1_FUNCTIONS(TSSPRIVKEY);
...
+ * @retval 0 on failure
+ */
+int
+init_auth(TPM2B_AUTH *a, const char *p, size_t l)
+{
+    if (!p) {
+        a->size = 0;
+        return 1;
+    }
+    if (0 == l) {
```
williamcroberts, line 126: no yoda checks, if zero is l is awkward, if l is zero is better.
```C
if (l == 0) {
```

src/tpm2-tss-engine-common.c
```diff
@@ -55,6 +56,84 @@ IMPLEMENT_ASN1_FUNCTIONS(TSSPRIVKEY);
...
+ * encoded contents. String interpretation can be controlled using prefixes as follows:
+ *
+ * * no prefix - Default to string interpretation.
+ * * `hex:` - Specify password in hex-string format.
+ * * `str:` - Force string interpretation, i.e. if the password starts with "hex:" or "str:".
+ * @param a The auth structure.
+ * @param p The password data.
+ * @param l The length of the password data (or 0 if NULL terminated).
+ * @retval 1 on success
+ * @retval 0 on failure
```
williamcroberts, line 117: if you're gonna use 1 or 0 as return values with 1 being success, just change the return type to bool and use stdbool true/false.

test/error_tpm2-tss-engine-common.c
```diff
@@ -21,7 +21,83 @@ __wrap_Esys_Initialize()
...
+void
+check_init_auth(void **state)
+{
+    (void)(state);
+    int i;
+    TPM2B_AUTH auth = { .size = -1 };
+
+    i = init_auth(&auth, NULL, 0);
+    assert_int_equal(i, 1);
+    assert_true(auth.size == 0);
```
williamcroberts, line 34: what about:
```C
init_auth(&auth, "", 0);
```

test/error_tpm2-tss-engine-common.c
```diff
@@ -21,7 +21,83 @@ __wrap_Esys_Initialize()
...
+
+
```
williamcroberts, line 36: extra newline, delete

test/error_tpm2-tss-engine-common.c
```diff
@@ -21,7 +21,83 @@ __wrap_Esys_Initialize()
...
+    assert_int_equal(i, 1);
+    assert_true(auth.size == 12);
+    assert_memory_equal(auth.buffer, "hex:DEADBEEF", auth.size);
+
+    i = init_auth(&auth, "str:hex:DEADBEEF", 10);
+    assert_int_equal(i, 1);
+    assert_true(auth.size == 10);
+    assert_memory_equal(auth.buffer, "str:hex:DE", auth.size);
+
+
```
williamcroberts, line 83: extra newline, delete

test/error_tpm2-tss-engine-common.c
```diff
@@ -21,7 +21,83 @@ __wrap_Esys_Initialize()
...
+    assert_int_equal(i, 1);
+    assert_true(auth.size == 7);
+    assert_memory_equal(auth.buffer, "str:str", auth.size);
+
+    i = init_auth(&auth, "str:hex:DEADBEEF", 0);
+    assert_int_equal(i, 1);
+    assert_true(auth.size == 12);
+    assert_memory_equal(auth.buffer, "hex:DEADBEEF", auth.size);
+
+    i = init_auth(&auth, "str:hex:DEADBEEF", 10);
```
williamcroberts, line 78: mix up the strings a bit, you only try deadbeef here and throughout.

Also, what about over-length passwords?

AndreasFuchsSIT:
@khwaaj Any updates here ?

khwaaj:
No updates yet, I've been working on non-TPM related projects for quite some time and I actually forgot about this PR. Appropriately I'm back to doing some TPM related work now, so I might find some time to look into this over the coming weeks.

AndreasFuchsSIT:
I'll close it for now since we had no updates for 6 months

PR #144, opened 2019-09-24
Bugfix of error messages at ECC code.

Baalmor:
This is a fix of probably, 'copy-paste' errors at ECC-related code. A error message if happened using ECC operations would lead to RSA instead.

Signed-off-by: Felix P <baalmor@gmail.com> Security researcher


diabonas:
Your signed-off tag seems a bit off since it doesn't match the commit author (which is probably also why [DCO bot complains](https://github.com/tpm2-software/tpm2-tss-engine/pull/144/checks?check_run_id=234109853)). Other than that LGTM.

PR #148, opened 2019-10-16
tpm2tss_tpm2data_readtpm(): be smarter about guessing emptyAuth.

asac:
With this patch tpm2tss_tpm2data_readtpm will assume empty Auth not only
if it has explicitely tested it for NODA objects, but also if it finds
an empty authPolicy in public areas.

With this change using tpm2_create'ed objects with default attributes
grows convenience as it does not ask for password if no auth policy is set.

Signed-off-by: Alexander Sack <asac@pantacor.com>

### DIFF
--- a/src/tpm2-tss-engine-common.c
+++ b/src/tpm2-tss-engine-common.c
@@ -243,10 +243,16 @@ tpm2tss_tpm2data_readtpm(uint32_t handle, TPM2_DATA **tpm2Datap)
         goto error;
     }
 
-    /* If the persistent key has the NODA flag set, we check whether it does
-       have an empty authValue. If NODA is not set, then we don't check because
-       that would increment the DA lockout counter */
-    if ((outPublic->publicArea.objectAttributes & TPMA_OBJECT_NODA) != 0) {
+    /* If the TPMT_PUBLIC area has a zero length AuthPolicy we go with emptyAuth */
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
authPolicy and authValue are completely different things. The engine currently makes no use of authPolicy anyways. The authValue is in the private area, not the public area.

Using this patch will increment the da-counter and may lead to lockout...
Sorry...

asac:
@AndreasFuchsSIT you are of course right. sorry for me being confused. In the end i am indeed after PCR policy support for tpm2tss engine; have there been patches, WIP things at some point that we could lean on to get a good start? Otherwise, could you outline the general approach we should take to introduce such support so I can see if we can help getting a first iteration going? Thanks for your input!

AndreasFuchsSIT:
Once FAPI is in place in tpm2-tss, we can leverage FAPI keys from here and have full policy support...
________________________________

asac:
ok cool. Is ETA close for FAPI in tpm2-tss? Do you know @AndreasFuchsSIT ?

AndreasFuchsSIT:
There's an RFC-PR open...
We need to do some more coding, but you can have a look and play around:
https://github.com/tpm2-software/tpm2-tss/pull/1537

________________________________

Issue #151, opened 2019-11-15
What if tpm2-tss was configured with "--disable-tcti-mssim"


jhealyt:
I haven't been able to confirm this is a problem for everyone (because I'm using my own Makefile), however I'll share my experience anyway:

It seems as though the key generator program "tpm2tss-genkey" depends upon the file "libtss2-tcti-mssim.so", however this file won't exist if tpm2-tss was configured with  "--disable-tcti-mssim".

AndreasFuchsSIT:
I wonder if the `-ltss2-tcti-mssim -ltss2-tcti-device` in the Makefile are actually needed anymore, since this is handled by tpm2-tss-engine-tcti.c anyways. Might just work without.

Also we really need to switch to libtss2-tctildr asap... :-(

jhealyt:
I've looked through the code to try figure this out before asking you. . .

. . . . but what is "tctildr" actually use for? I notice that I have another library file "/usr/lib/libtss2-tctildr.so"

Is it some sort of dynamic loader of other TCTI's ?

AndreasFuchsSIT:
Exactly

diabonas:
> I wonder if the `-ltss2-tcti-mssim -ltss2-tcti-device` in the Makefile are actually needed anymore, since this is handled by tpm2-tss-engine-tcti.c anyways. Might just work without.

These shouldn't be necessary, the TCTI libraries are `dlopen()`'ed, so they don't need to be linked. I can have a look a porting the code to tss2-tctildr, shouldn't be much more that removing [`tpm2-tss-engine-tcti.c`](https://github.com/tpm2-software/tpm2-tss-engine/blob/master/src/tpm2-tss-engine-tcti.c) and replacing the custom `tcti_` functions in [`tpm2-tss-engine.c`](https://github.com/tpm2-software/tpm2-tss-engine/blob/master/src/tpm2-tss-engine.c) by their tctildr equivalents.

PR #153, opened 2019-11-21
Use tss2-tctildr

diabonas:
tss2-tctildr encapsulates handling of the TCTI context, allowing us to remove our own TCTI management. It is available in tpm2-tss 2.3.

For easier review, the PR is split into three commits: the first commit fixes the call order of `esys_auxctx_free`, see the [review comment](https://github.com/tpm2-software/tpm2-tss-engine/pull/153#discussion_r349096404). The second commit (which is functional on its own) removes [`tpm2-tss-engine-tcti.c`](https://github.com/tpm2-software/tpm2-tss-engine/blob/fdc8f65dfc8bad8b5a3aed181fae338267308f70/src/tpm2-tss-engine-tcti.c) and replaces it by the corresponding tss2-tctildr functions. It still uses the custom [`ESYS_AUXCONTEXT` structure](https://github.com/tpm2-software/tpm2-tss-engine/blob/fdc8f65dfc8bad8b5a3aed181fae338267308f70/src/tpm2-tss-engine-common.h#L65-L68), which encapsulates the ESYS context as well as a `dlopen()` handle for the TCTI library that is not needed any more after the changes in this commit. Therefore the third commit uses the normal `ESYS_CONTEXT` provided by tpm2-tss instead, which is mostly boring search and replace of function and variable names.

Fixes: #151


src/tpm2-tss-engine-common.c
```diff
@@ -57,66 +57,63 @@ IMPLEMENT_ASN1_FUNCTIONS(TSSPRIVKEY);
...
-        r = Esys_GetTcti(eactx_p->ectx, &tcti_ctx);
+        r = Esys_GetTcti(*esys_ctx, &tcti_ctx);
         if (TSS2_RC_SUCCESS != r) {
-            ERR(esys_auxctx_free, TPM2TSS_R_GENERAL_FAILURE);
+            ERR(esys_ctx_free, TPM2TSS_R_GENERAL_FAILURE);
         } else {
             Tss2_TctiLdr_Finalize(&tcti_ctx);
         }
-        Esys_Finalize(&(eactx_p->ectx));
+        Esys_Finalize(esys_ctx);
```
AndreasFuchsSIT: This was obviously a bug in the previous code already, but we want to call Esys_Finalize() before Tss2_TctiLdr_Finalize() in case of Esys_Finalize still wanting to perform some actions on TCTI.

diabonas: Sure, I already changed the logic to always call `Esys_Finalize` regardless of whether `Esys_GetTcti` succeeded or not, so I don't mind fixing this as well, I added a new commit for these call order fixes. I was worried about a potential use after free first, but `Esys_Finalize` never touches the TCTI given by the user, it is just passed around as a pointer, so this is fine.

diabonas:
I forgot to remove the (already obsolete) linking of `tpm2tss-genkey` against `tss2-tcti-mssim` and `tss2-tcti-device`. Fixed now, and corresponding bug report referenced in the PR description.

PR #182, opened 2020-07-03
Support ossl engine config

williamcroberts:
The library has two methods of controling the TCTI, one is via an env variable, the other is through the engine control interface. Unify these so the tpm2tss-genkey will support eithir method.

This patch series adds support for openssl config to be understood by this tool as well as exposing the tcti option via -t to override the config file. It also changes the other options to override only when set.

### DIFF
--- a/src/tpm2-tss-engine.c
+++ b/src/tpm2-tss-engine.c
@@ -270,7 +270,9 @@ init_engine(ENGINE *e) {
 #ifdef ENABLE_TCTIENVVAR
     /*  Set the default TCTI option from the environment */
     OPENSSL_free(tcti_nameconf);
-    tcti_nameconf = OPENSSL_strdup(getenv("TPM2TSSENGINE_TCTI"));
+    if (getenv("TPM2TSSENGINE_TCTI")) {
+        tcti_nameconf = OPENSSL_strdup(getenv("TPM2TSSENGINE_TCTI"));
+    }
 #endif
 
     rc = init_rand(e);
--- a/src/tpm2-tss-engine.c
+++ b/src/tpm2-tss-engine.c
@@ -28,6 +28,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
  * THE POSSIBILITY OF SUCH DAMAGE.
  ******************************************************************************/
+#include "config.h"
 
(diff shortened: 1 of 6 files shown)


src/tpm2-tss-engine.c
```diff
@@ -270,7 +271,9 @@ init_engine(ENGINE *e) {
 #ifdef ENABLE_TCTIENVVAR
     /*  Set the default TCTI option from the environment */
     OPENSSL_free(tcti_nameconf);
-    tcti_nameconf = OPENSSL_strdup(getenv("TPM2TSSENGINE_TCTI"));
+    if (getenv("TPM2TSSENGINE_TCTI")) {
+        tcti_nameconf = OPENSSL_strdup(getenv("TPM2TSSENGINE_TCTI"));
+    }
```
williamcroberts, line 276: This hunk here is pretty important, it segfaults if it's ENABLE_TCTIENVVAR is defined but the env var TPM2TSSENGINE_TCTI is not set.... I can split this out to a separate PR if you want to take it but need changes elsewhere in this PR

williamcroberts: The reason, at least as seen in Ubuntu 16.04 and its version of OSSL is that OPENSSL_strdup tries to access the NULL pointer.

AndreasFuchsSIT:
LGTM, but does config.h.in need to be part of repo ? Ain't that generated by bootstrap ?

williamcroberts:
> config.h.in

Right you are, I need to drop that from the other projects I added it in.

williamcroberts:
@tstruk  or @AndreasFuchsSIT Build died building OSSL dependency have you been seeing:
```bash
rm -f test/enginetest
${LDCMD:-gcc} -pthread -m64 -Wa,--noexecstack -Wall -O3 -L.   \
	-o test/enginetest test/enginetest.o \
	 test/libtestutil.a -lcrypto -ldl -pthread 
test/ectest.o: In function `custom_generator_test':
ectest.c:(.text+0x12e): undefined reference to `EC_GROUP_get_field_type'
ectest.c:(.text+0x22b): undefined reference to `EC_GROUP_get0_field'
collect2: error: ld returned 1 exit status
Makefile:8616: recipe for target 'test/ectest' failed
make[1]: *** [test/ectest] Error 1
```

PR #185, opened 2020-08-10
engine-common: drop memset zero and use zalloc

williamcroberts:
Rather than OPENSSL_malloc + memset of 0, use OPENSSL_zalloc and do it
in less code and possibly more efficiently.

### DIFF
--- a/src/tpm2-tss-engine-common.c
+++ b/src/tpm2-tss-engine-common.c
@@ -209,12 +209,11 @@ tpm2tss_tpm2data_readtpm(uint32_t handle, TPM2_DATA **tpm2Datap)
     ESYS_CONTEXT *esys_ctx = NULL;
     TPM2B_PUBLIC *outPublic;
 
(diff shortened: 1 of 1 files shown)


src/tpm2-tss-engine-common.c
```diff
@@ -209,12 +209,11 @@ tpm2tss_tpm2data_readtpm(uint32_t handle, TPM2_DATA **tpm2Datap)
     ESYS_CONTEXT *esys_ctx = NULL;
     TPM2B_PUBLIC *outPublic;
 
-    tpm2Data = OPENSSL_malloc(sizeof(*tpm2Data));
+    tpm2Data = OPENSSL_zalloc(sizeof(*tpm2Data));
```
williamcroberts, line 212: Openssl 1_0_2g does not have this... on top of that, thier zalloc is literally just malloc + memset anyways. dropping

PR #193, opened 2020-09-30
test: add support for swtpm simulator

diabonas:
This is a straightforward port of https://github.com/tpm2-software/tpm2-totp/pull/69 to add support for swtpm, since both project share almost the same codebase for the integration test framework.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,10 @@ dist: xenial
 
 env:
   matrix:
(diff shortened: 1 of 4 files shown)


AndreasFuchsSIT (changes requested):
I think we should switch CI over to swtpm in the same step ?

diabonas:
> I think we should switch CI over to swtpm in the same step ?

Sure, added as a separate commit to this PR.

PR #140, opened 2019-08-28
Add: import of TPM keys

t-higuchi:
I have merged https://github.com/AndreasFuchsSIT/tpm2-tss-engine/commit/7f35306d7374940ff51f349965b67b83b4dabfd4 , removed -j option, and fixed testcases so that they works with tpm2-tools 3.2.1-rc0.

I confirmed that existing openssl-generated RSA keys can be imported with tpm2-tools-4.0-rc0 in this way:

`tpm2_createprimary -C o -g sha256 -G ecc -a "fixedtpm|fixedparent|sensitivedataorigin|noda|decrypt|restricted|userwithauth"  -c primary.ctx`
`tpm2_import -G rsa -g sha256 -i rsakey.pem -C primary.ctx -u tpmkey.pub -r tpmkey.priv`
`tpm2tss-genkey -i tpmkey.pub -k tpmkey.priv  rsakey.tss`

I did not added testcase for this though since it requires migration to tpm2-tools-4.0, which breaks other testcases.
so Issue #39 could be closed after migration to tpm2-tools-4.0.

Issue #39, opened 2018-10-24
Import existing keys into TPM for use with openssl


geobra:
Hi

My usecase is to import existing keys into the TPM to be able to use it for an TLS connection using openssl.

What works so far is to generate a key inside tpm on client and server, generate the certificates and connect:

-> execute on client or server accordingly...
"export TYPE=client|server"

# generate key in tpm
tpm2tss-genkey -a rsa $TYPE.tss

# gen a cert sign request
openssl req -new -key $TYPE.tss -engine tpm2tss -keyform engine -out $TYPE.csr

# gen the certificate
openssl x509 -req -in $TYPE.csr -CA ca-root.pem -CAkey ca-key.pem -CAcreateserial -out $TYPE-cert.pem -days 365

on server:
openssl s_server -cert server-cert.pem -key server.tss -keyform engine -engine tpm2tss -CAfile ca-root.pem -accept 8443 -Verify 1

on client:
openssl s_client -keyform engine -engine tpm2tss -connect <server>:8443 -cert client-cert.pem -key client.tss -CAfile ca-root.pem

But what I really want is to use existing private key and certificate on the client. So I tried to import my key into the TPM.

# generate primary
tpm2_createprimary -G 0x0001 -g 0x000b -H o -C parent.ctx
tpm2_import -C parent.ctx -G 0x0001 -k client-key.pem -u client-key-tpm.pub -r client-key-tpm.priv 

But now I have the encrypted parts of the private and public keys in a binary format. I need it in tss format for openssl to be usefull.

Can this be converted? 
Where to go from here? Or am I on the wrong path with this approach?

AndreasFuchsSIT:
This is a duplicate of #12 though title and description are better.
It's a planned feature...

dwmw2:
And yes... in the meantime, assuming that primary key you generated is the same as the one the engine would generate, you can actually take those binary blobs and put them into a PEM file. How are you at writing ASN.1 structures by hand...? :)

Run `openssl asn1parse -in client tss` then replace the OCTET-STRINGs for the pub and priv key with the blobs you have...

geobra:
Thanks for this information! I am off for one week now and will try this when I am back.

AndreasFuchsSIT:
Note for implementer: This is about importing keys generated with the tpm2-tools

williamcroberts:
[...]

PR #194, opened 2020-10-24
TPM key import

gotthardp:
May I help finalizing the import feature (#39), please? I took over the content of the #140 and fixed (most of the) review comments, except the following.

Pending questions:
"Can't we just import a public key only?" (@williamcroberts) -- Right now the tpm2tss-genkey generates both public and private portion. Consistently with this, the import function lets the user to import public&private portion of a pre-existing key.

"Shouldn't we make the arguments make consistent with tpm2-tools?" (@gotthardp) The tpm2-tools consistently use `-u` for the public portion and `-r` for the private portion. I'd suggest to stick to this convention and use -u/-r instead of -i/-k. Would you agree?

Is there anything else that should be addressed?

### DIFF
--- a/man/tpm2tss-genkey.1.md
+++ b/man/tpm2tss-genkey.1.md
@@ -1,6 +1,6 @@
 % tpm2tss-genkey(1) tpm2-tss-engine | General Commands Manual
 %
-% JUNE 2018
+% OCTOBER 2020
 
 # NAME
 **tpm2tss-genkey**(1) -- generate TPM keys for tpm2-tss-engine
@@ -31,6 +31,13 @@ key information. This file can then be loaded with OpenSSL using
 
   * `-c <curve>`, `--curve <curve>`:
     If alg ecdsa is chosen, the curve for ecc (default: nist_p256)
+
+  * `-u <file>`, `--public <file>`:
+    Public key (TPM2B_PUBLIC) to be imported. Requires `-r`.
+
+  * `-r <file>`, `--private <file>`:
+    The (encrypted) private key (TPM2B_PRIVATE) to be imported.
+    Requires `-u`.
 
   * `-e <exponent>`, `--exponent <exponent>`:
     If alg rsa is chosen, the exponent for rsa (default: 65537)
--- a/src/tpm2tss-genkey.c
+++ b/src/tpm2tss-genkey.c
@@ -55,6 +55,8 @@ char *help =
     "    -c, --curve     curve for ecc (default: nist_p256)\n"
     "    -e, --exponent  exponent for rsa (default: 65537)\n"
     "    -h, --help      print help\n"
+    "    -u, --public    import a key and read its public portion from this file\n"
+    "    -r, --private   import the sensitive key portion from this file\n"
     "    -o, --ownerpw   password for the owner hierarchy (default: none)\n"
     "    -p, --password  password for the created key (default: none)\n"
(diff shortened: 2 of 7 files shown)


AndreasFuchsSIT:
Awesome ! Thanks for taking up this task.

I'd be fine with using -u and -r for the keyfiles.

Please also sQuash both commits into one with you as the author. You can add the other authors using a `Co-authored-by` tag.
And also please add some explanatory commit message as well.

I'll be VERY happy to merge this then. :-)

man/tpm2tss-genkey.1.md
```diff
@@ -32,6 +32,12 @@ key information. This file can then be loaded with OpenSSL using
   * `-c <curve>`, `--curve <curve>`:
     If alg ecdsa is chosen, the curve for ecc (default: nist_p256)
 
+  * `-i <file>`, `--importpub <file>`:
+    Public key (TPM2B_PUBLIC) to be imported. Requires `-k`.
+
+  * `-k <file>`, `--importtpm <file>`:
+    The (encrypted) private key (TPM2B_PRIVATE) to be imported.
```
williamcroberts, line 39: doesn't `-k` need `-i` as well?

Change `-i` to `-u` and `-k` to `-r` to be consistent with tpm2-tools.

williamcroberts (changes requested):

src/tpm2tss-genkey.c
```diff
@@ -207,6 +221,12 @@ parse_opts(int argc, char **argv)
         ERR("%s", help);
         exit(1);
     }
+
+    if ((opt.importpub && !opt.importtpm) || (!opt.importpub && opt.importtpm)) {
+        ERR("--public requires --private");
```
williamcroberts, line 226: The message depends on which one is missing, so you can get the wrong message here.

```C
/* XXX I think this will do it */
if (opt.importpub != opt.importpub) {
   ERR (opt.importpub ? "--public requires --private" :
       "--private requires --public");
}

**Review by williamcroberts — CHANGES_REQUESTED** [review]

File: src/tpm2tss-genkey.c (hunk 2 above)
Comment (gotthardp) [reply src/tpm2tss-genkey.c:226]: The opt.importpub is a char pointer-- the (opt.importpub != opt.importpub) will be always true, unless both are NULL. We could either have two conditions, or change the error message to say, e.g.:
```C
    if ((opt.importpub && !opt.importtpm) || (!opt.importpub && opt.importtpm)) {
        ERR("Import requires both --public and --private");
        return 1;
    }
```

gotthardp: I just pushed another revision that somehow combines both approaches.

williamcroberts (approved):

src/tpm2tss-genkey.c
williamcroberts: > I just pushed another revision that somehow combines both approaches.

glad you sorted it out, yeah I forgot to convert them to boolean values, but I see you nailed it with the ``!!``.

williamcroberts:
LGTM, we'll see what CI has to say and if Andreas has anymore feedback.

AndreasFuchsSIT (approved):
LGTM, Thanks a lot !

williamcroberts:
I used this to test GH Actions since it was an approved PR. Its been moved over to here and merged:
  - https://github.com/tpm2-software/tpm2-tss-engine/pull/202

PR #207, opened 2021-02-27
Add basic implementation for ECDH.

### LINKED ISSUE #206: Add ECDH support if compatible with OpenSSL ECDH scheme.
Issue1: you must provide a valid hash length input. I.e. you need to provide 20 or 32 bytes as input.
Then it should work. Or you let OpenSSL perform a hashing operation first.

Issue2: The engine currently does not support ECDH. I don't know if TPM's ECDH scheme and OpenSSL's ECDH scheme are compatible to begin with. That would need to be investigated.

_Originally posted by @AndreasFuchsSIT in https://github.com/tpm2-software/tpm2-tss-engine/issues/178#issuecomment-633492417_

abeltrano:
Fixes #206.

### Add support for ECDH

The implementation uses TPM2_ECDH_ZGen as provided by tss2-esys ([Esys_ECDH_ZGen()](https://github.com/tpm2-software/tpm2-tss/blob/ac2011c715f8611c44d073efc059fad4d3e7ed77/src/tss2-esys/api/Esys_ECDH_ZGen.c#L62)). A persistent handle value referencing the TPM private key is specified as the key context. Like other engine function implementations, a password session (ESYS_TR_PASSWORD) is used for authorization of the session handle. The public point is expected in uncompressed form.

**ECDH Validation**
Use newly added integration test under tests/ecdh.sh.

### DIFF
--- a/src/tpm2-tss-engine-ecc.c
+++ b/src/tpm2-tss-engine-ecc.c
@@ -414,7 +414,11 @@ tpm2tss_ecc_makekey(TPM2_DATA *tpm2Data)
  * @retval NULL on failure.
  */
 TPM2_DATA *
+#if OPENSSL_VERSION_NUMBER < 0x10100000
 tpm2tss_ecc_getappdata(EC_KEY *key)
+#else /* OPENSSL_VERSION_NUMBER < 0x10100000 */
+tpm2tss_ecc_getappdata(const EC_KEY *key)
+#endif /* OPENSSL_VERSION_NUMBER < 0x10100000 */
 {
     if (ec_key_app_data == -1) {
         DBG("Module uninitialized\n");
--- a/src/tpm2-tss-engine-ecc.c
+++ b/src/tpm2-tss-engine-ecc.c
@@ -117,7 +117,102 @@ static int EC_GROUP_order_bits(const EC_GROUP *group)
     BN_free(order);
     return ret;
 }
-#endif
+#else /* OPENSSL_VERSION_NUMBER < 0x10100000 */
+ /**
+  * Initialize a TPM2B_ECC_POINT from an OpenSSL EC_POINT.
+  *
+  * @param point Pointer to output tpm point
+  * @param pub_key OpenSSL public key to convert
+  * @param group Curve group
+  * @retval 0 on failure
+  */
+static int
+init_tpm_public_point(TPM2B_ECC_POINT *point, const EC_POINT *ec_point,
+                        const EC_GROUP *ec_group)
+{
+    unsigned char buffer[1 + sizeof(point->point.x.buffer)
+                           + sizeof(point->point.y.buffer)];
+    BN_CTX *ctx = BN_CTX_new();
+    if (!ctx)
+        return 0;
+
+    BN_CTX_start(ctx);
+    size_t len = EC_POINT_point2oct(ec_group, ec_point,
+                    POINT_CONVERSION_UNCOMPRESSED, buffer, sizeof(buffer), ctx);
+    BN_CTX_free(ctx);
+    len = (len - 1) / 2;
+
+    point->point.x.size = len;
+    point->point.y.size = len;
+    memcpy(point->point.x.buffer, &buffer[1], len);
+    memcpy(point->point.y.buffer, &buffer[1 + len], len);
+
+    return 1;
+}
+
+/**
+ * Generate a shared secret using a TPM key
+ *
+ * @param psec Pointer to output buffer holding shared secret
+ * @param pseclen Size of the psec buffer
+ * @param pub_key The peer's public key
+ * @param ecdh The ECC key object for the host private key
+ * @retval 0 on failure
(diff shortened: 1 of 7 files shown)


AndreasFuchsSIT (changes requested):
Could you sQuash the first two and the last two commits ?
The first for consistency and the last to so we're always runnable.

Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.

abeltrano:
> Could you sQuash the first two and the last two commits ?
> The first for consistency and the last to so we're always runnable.
> 
> Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.

Sure, no problem.

abeltrano:
> Could you sQuash the first two and the last two commits ?
> The first for consistency and the last to so we're always runnable.
> 
> Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.



> > Could you sQuash the first two and the last two commits ?
> > The first for consistency and the last to so we're always runnable.
> > Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.
> 
> Sure, no problem.
I've added an integration test for ecdh but not yet enabled it. Since this feature requires openssl >= 1.1.x, is there a facility in place to skip tests for certain ossl versions?

williamcroberts (commented):
@idesai remember when we wanted to see what ECDH would look like end-to-end, well this appears to be it. This will give us a good model to follow for tpm2-tools and tpm2-pkcs11.

abeltrano:
> > Could you sQuash the first two and the last two commits ?
> > The first for consistency and the last to so we're always runnable.
> > Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.
> 
> > > Could you sQuash the first two and the last two commits ?
> > > The first for consistency and the last to so we're always runnable.
> > > Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.
> > 
> > 
> > Sure, no problem.
> 
> I've added an integration test for ecdh but not yet enabled it. Since this feature requires openssl >= 1.1.x, is there a facility in place to skip tests for certain ossl versions?

Tests are now conditionally run based on presence of the OpenSSL API required. My autoconf-fu isn't the best, but it works. @AndreasFuchsSIT, should be ready for your final review now.

abeltrano:
> > > Could you sQuash the first two and the last two commits ?
> > > The first for consistency and the last to so we're always runnable.
> > > Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.
> > 
> > 
> > > > Could you sQuash the first two and the last two commits ?
> > > > The first for consistency and the last to so we're always runnable.
> > > > Also the tests are actually just bash scripts, so the example from the PR text could directly be used there. I don't want to merge without a test and need some time before writing one myself.
> > > 
> > > 
> > > Sure, no problem.
> > 
> > 
> > I've added an integration test for ecdh but not yet enabled it. Since this feature requires openssl >= 1.1.x, is there a facility in place to skip tests for certain ossl versions?
> 
> Tests are now conditionally run based on presence of the OpenSSL API required. My autoconf-fu isn't the best, but it works. @AndreasFuchsSIT, should be ready for your final review now.

@AndreasFuchsSIT please let me know if any additional changes are needed to complete this PR. Thanks.

src/tpm2-tss-engine-ecc.c
```diff
@@ -595,6 +694,7 @@ init_ecc(ENGINE *e)
         = NULL;
     EC_KEY_METHOD_get_sign(ecc_methods, &orig_sign, NULL, NULL);
     EC_KEY_METHOD_set_sign(ecc_methods, orig_sign, NULL, ecdsa_sign);
+    EC_KEY_METHOD_set_compute_key(ecc_methods, ecdh_compute_key);
```
AndreasFuchsSIT, line 697: I think this needs to be guarded by an #ifdef as well, right ?

AndreasFuchsSIT (changes requested):

src/tpm2-tss-engine-ecc.c
abeltrano: This is defined within `#else /* OPENSSL_VERSION_NUMBER < 0x10100000 */` on line 697, which I think is sufficient. Also, the tests pass which build for all supported versions of OpenSSL, so I think it's good as-is. Though, I defer to your experience in the codebase here.

AndreasFuchsSIT: Oh, sorry. I missed that.
Cool... I'll merge it !

AndreasFuchsSIT (approved):

Issue #161, opened 2020-02-07
Possible memleak in tpm2tss_rsa_makekey


lcuminato:
Hi,

I'm trying to assess if a possible memory leak exists in `tpm2tss_rsa_makekey`. 
The following code stores the `tpm2Data` object into the `rsa` object by using this macro `RSA_set_app_data`
```
if (!RSA_set_app_data(rsa, tpm2Data)) {
        ERR(populate_rsa, TPM2TSS_R_GENERAL_FAILURE);
        goto error;
    }
```
I think the problem here is that when the application code is doing cleanup and calls `EVP_PKEY_free` to destroy the `EVP_PKEY` object, which will then eventually call RSA_free() to free the `RSA` object, `RSA_free()` is not aware of the `tpm2Data` object and therefore it won't deallocate it.

I figure this could be it by running valgrind on my application, which is currently leaking memory at a rate of 2kb per CURL call (which uses the tpm2 as a OpenSSL engine).

This is my valgrind report:

```
==15907== 173,096 bytes in 77 blocks are definitely lost in loss record 58 of 58
==15907==    at 0x4A07ECB: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==15907==    by 0x31FBD8DED7: CRYPTO_malloc (in /usr/lib/libcrypto.so.1.0.2)
==15907==    by 0xCB0E292: ???
==15907==    by 0xCB0D7B5: ???
==15907==    by 0x31FBD0E15C: ENGINE_load_private_key (in /usr/lib/libcrypto.so.1.0.2)
==15907==    by 0x31FB849E35: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB84A2DF: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB84B086: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB80F6E1: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB810EDA: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB82BC2F: ??? (in /usr/lib/libcurl.so.4.6.0)
==15907==    by 0x31FB82CEDA: curl_multi_perform (in /usr/lib/libcurl.so.4.6.0)
```
gdb:
```
[...]

PR #217, opened 2021-08-06
fix: mem leak in loadkey function

zaporozhets:
This patch fixes mem leak in `loadkey()` function.

Memory(`tpm2Data`) allocated by `tpm2tss_tpm2data_readtpm`/`tpm2tss_tpm2data_read` remains not freed.


zaporozhets:
Just found that under the hood it uses `RSA_set_app_data` to keep `tpm2Data`. So seems that memleak in my case comes from `RSA_free`.(might be similar to https://github.com/tpm2-software/tpm2-tss-engine/issues/161)

AndreasFuchsSIT:
> Just found that under the hood it uses `RSA_set_app_data` to keep `tpm2Data`. So seems that memleak in my case comes from `RSA_free`.(might be similar to #161)

Does that mean you want to change something or is the fix still valid ?

platonef:
@AndreasFuchsSIT The fix is added in the loadkey function where the keys are created. 
Is that function the correct place to free the tpm2Data? That is, is the tpm2Data used anywhere else, and if they are freed in the loadkey function will create a problem? 

In this #161, the deallocation of the tpm2Data is added in the finish function. Should be done the same here? 

Thus, which do you think is the best way to free the tpm2Data?

Issue #221, opened 2021-09-03
Specifying parent with -P flag in tpm2tss-genkey yields overflow


kuhlmannmarkus:
Hi all,

Im using `tpm2tss-genkey` to derive a PEM representation from some key material I generated via https://github.com/google/go-attestation. Therefore I export the public and private portions according to https://github.com/google/go-tpm/issues/233 and then do `tpm2tss-genkey -u pubktpmmarshal -r pktpmmarshal -P 0x81000001 go-key.pem`. The key generally seems fine, but I found that the parent part probably has an overflow. `cat go-key.pem | openssl asn1parse` yields:

>    0:d=0  hl=4 l= 498 cons: SEQUENCE
>    4:d=1  hl=2 l=   6 prim: OBJECT            :2.23.133.10.1.3
>   12:d=1  hl=2 l=   3 cons: cont [ 0 ]
>   14:d=2  hl=2 l=   1 prim: BOOLEAN           :1
>   **17:d=1  hl=2 l=   4 prim: INTEGER           :-7EFFFFFF**
>   23:d=1  hl=4 l= 280 prim: OCTET STRING      [HEX DUMP]:01160001000B0004007200000010001008000000000001009203AB176AD723B0B51BBE717115558F5AFC48A8171F81E4925CF3212E9915C506F6711ECF4957BDE2DBD2EE26856776083FBC8A26BCA37C73392E4CBC5E5F7C9FCEC13AED9EF7F39B7CC18A95EA1BFE7EE95BA7FB1CEF9CE42B07974DE0DC23A89FCF603A1523691FF877DD8F3623C6A2FE1C506CCDA2530C99965C36BD0C5CD43C750BE3F7F6772387F4779A150E427C843CDA73C6575EABA53E6FF1E3A908EC7D6B3BF78F45B67C7C814F11F863E2E40D04DC0E02E9D49A6C265C236B7DB77064037F5D9D5007DF56AB1D45A1EF709274F27A6AADA0CAA1DD0D865115775EBAFD5BEA12AD48125F364210691671F89C070FB6A405249533567BE35654FAA7
[...]

PR #222, opened 2021-09-07
use BIGNUM for parent handle

whooo:
This will encode/decode the parent handle as an unsigned long on 32bit systems.

Fixes https://github.com/tpm2-software/tpm2-tss-engine/issues/221


whooo:
I just need to find a good way to handle keys with an negative parent handle so nothing breaks

whooo:
Will leave the decoding as is as it seems to work fine with negative parent handles, even on 64 bit systems

AndreasFuchsSIT:
Just to make sure: Does this still work with PEM files generated by the old implementation ?

whooo:
> Just to make sure: Does this still work with PEM files generated by the old implementation ?

According to my testing, I was able to use a key with -7EFFFFFF as the parent handle with openssl pkeyutl -sign, but I was only able to test the openssl parts on a 64bit system

williamcroberts:
Dont worry about LGTM fix over here: https://github.com/tpm2-software/tpm2-tss-engine/pull/223

williamcroberts:
We could 

> > Just to make sure: Does this still work with PEM files generated by the old implementation ?
> 
> According to my testing, I was able to use a key with -7EFFFFFF as the parent handle with openssl pkeyutl -sign, but I was only able to test the openssl parts on a 64bit system

We could set up the i386 container so we get 32bit coverage.

williamcroberts:
> 32Bit-Containers would be nice (for all tss projects I guess).

It's available now and used by tpm2-abmrd. However it's arm32v7 not i386. It should just be as simple as adding it to your container list (done here https://github.com/tpm2-software/tpm2-tss-engine/pull/224)

> But I'm mostly concerned about backwards-compatibility.
> Maybe we could add a test that explicitely loads one of those old PEM files with the negative parent handle ?

Good idea, I can take one and ensure that tpm2_ptool can handle it.

williamcroberts:
> 32Bit-Containers would be nice (for all tss projects I guess).
> But I'm mostly concerned about backwards-compatibility.
> Maybe we could add a test that explicitely loads one of those old PEM files with the negative parent handle ?

Is that actually going to be possible because the priv/pub blobs will be for differing TPMs?

whooo:
I could add a unit test for tpm2tss_tpm2data_read using an already generated key with a negative parent handle.

AndreasFuchsSIT:
Perfect

williamcroberts (dismissed):
LGTM ill let @AndreasFuchsSIT give the final sign-off

williamcroberts:
> LGTM ill let @AndreasFuchsSIT give the final sign-off

Oh let me "approve" the CI run, so obviously that needs to pass.

williamcroberts:
> I'm unsure on why the unit test fails, works well on my laptop, is there anyway to get access to test-suite.log?

Just expand the failure tab, that has the output:
```
Run cat build/test-suite.log || true
=================================================
   tpm2-tss-engine 1.1.0-rc1: ./test-suite.log
=================================================

# TOTAL: 19
# PASS:  18
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test/tpm2-tss-engine-common
=================================

[==========] Running 1 test(s).
[ RUN      ] check_tpm2tss_tpm2data_read
[  ERROR   ] --- 0 != 0x1
[   LINE   ] --- ../test/tpm2-tss-engine-common.c:19: error: Failure!
[  FAILED  ] check_tpm2tss_tpm2data_read
[==========] 1 test(s) run.
[  PASSED  ] 0 test(s).
[  FAILED  ] 1 test(s), listed below:
[  FAILED  ] check_tpm2tss_tpm2data_read

 1 FAILED TEST(S)
FAIL test/tpm2-tss-engine-common (exit status: 1)
```

williamcroberts:
I don't understand why I don't see error messages, since every failure path seems to have an ERR message. However, I think its becuase we build in a variant directory `./bootstrap && mdir build && cd build && ../configure --enable-unit --enable-debug && make -j4`. So it's not finding the file path. If im right, you need to add srcdir to it. I would pass a `-D` in CFLAGS that has the path to the PEM file.

williamcroberts:
Oh you need to add that file to make dist so that file exists... Add to EXTRA_DIST.

AndreasFuchsSIT:
@williamcroberts Do we all agree on merging this now ?

AndreasFuchsSIT (approved):

AndreasFuchsSIT:
@whooo Sorry this took so long, and now I've even introduced a merge conflict,
Could you rebase to master and resolve it ?
Highly appreciated.
Afterwards, I'll roll the next stable release...

whooo:
> @whooo Sorry this took so long, and now I've even introduced a merge conflict, Could you rebase to master and resolve it ? Highly appreciated. Afterwards, I'll roll the next stable release...

Done

williamcroberts (approved):

whooo:
I'm not able to reproduce the issue with the multi-arch test, they work in a 32 bit chroot on a x86_64 system and I don't have access to any ARM device that I could run Linux on

whooo:
I had to redo the loading part, so it will check for a negative number and in that case it will use ASN1_INTEGER_get instead of BN_get_word.

AndreasFuchsSIT (approved):

PR #232, opened 2021-10-04
Fix key app data memory leaks

rshearman:
* ecc: Fix leak of app data
* rsa: Fix leak of app data


JuergenReppSIT:
In `tpm2tss_rsa_genkey` it shout be checked `(rsa != NULL)` before  ` RSA_set_app_data(rsa, NULL);` in the error case because the unit test fails with a segfault.

JuergenReppSIT:
> It seems to me that `tpm2tss_rsa_genkey` assumes that rsa is non-NULL when being called as there is nothing guarding against this before `RSA_set_app_data(rsa, tpm2Data)` is called, and nothing sets rsa to NULL only in the error case. Furthermore, I don't see any segfaults when I run the unit tests.

Sorry, as a maintainer (occasionally) I didn't realize that a maintainer has to approve the running of the tests. I tested your branch locally. Now I did approve the running of the tests and you can see the error caused by the unit test `check_init_tpm_parent_via_api`

PR #234, opened 2021-11-29
tpm2-tss-engine: add digest like sha1,sha256

aaronUniontech:

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,7 +73,8 @@ libtpm2tss_la_SOURCES = src/tpm2-tss-engine.c \
                         src/tpm2-tss-engine-err.h \
                         src/tpm2-tss-engine-ecc.c \
                         src/tpm2-tss-engine-rand.c \
-                        src/tpm2-tss-engine-rsa.c
+                        src/tpm2-tss-engine-rsa.c   \
+                        src/tpm2-tss-engine-digest.c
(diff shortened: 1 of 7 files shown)


AndreasFuchsSIT:
I'm sorry, but I do not see a reason to have the TPM calculate a digest.
Could you explain to me, why this would be an advantage ?
Because otherwise, I'd have to reject this PR

aaronUniontech:
> I'm sorry, but I do not see a reason to have the TPM calculate a digest. Could you explain to me, why this would be an advantage ? Because otherwise, I'd have to reject this PR

maybe all the work can be done by the hardwork.
yes, it's not such necessary.

williamcroberts:
> > I'm sorry, but I do not see a reason to have the TPM calculate a digest. Could you explain to me, why this would be an advantage ? Because otherwise, I'd have to reject this PR
> 
> maybe all the work can be done by the hardwork. yes, it's not such necessary.

The TPM is very slow, so in general unless you're working in a way where you need a ticket, theirs no point to using the TPM to hash over software.

AndreasFuchsSIT:
So I appreciate the effort of this implementation but I'll have to decline the PR.
I'm sorry, but there is no advantage, only disadvantages in calculating the digest inside the TPM.

PR #237, opened 2021-12-02
Add ability to digest & sign with restricted ECC key

rshearman:
The TPM architecture mandates that a restricted signing key may only
sign a digest that has been produced by the TPM. Having a TPM key be
restricted is useful for implementing attestation using the key.

The current mechanism for signing data using OpenSSL with an ECC key
is through the EC_KEY sign method. This is too low-level, since it
relies on the digest having already been done (in software). The EVP
pkey object is the right level, since OpenSSL exposes EVP_DigestSign*
functions that operate on it.
    
So implement methods on EVP pkey for digest and sign for the EC type,
inheriting other EVP pkey methods from the built-in EC type in
OpenSSL. The digest and sign operation is implemented by making use of
the TPM to perform the digest and then passing the validation ticket,
which is required for a restricted key, (along with digest data) into
the subsequent sign operation in the TPM.
    
Make use of the digest_custom method, rather than signctx_init method
for performing initialisation of our context used for the
digest-and-sign operation so that the message digest type is already
set in the OpenSSL message digest context. It isn't possible to
override the builtin methods registered for OpenSSL message digest
types in a general sense, but it is possible to override the update
method called during a updating digest data in the context of
EVP_DigestSign* functions, so this is what is done to allow the digest
update to be performed in the TPM.
    
Whilst this is only implemented for ECC keys currently, the
implementation is done in a way where the parts not specific to the
key type are in a file not specific to the key type, so that they can
be used for doing the same with RSA in the future.

### DIFF
--- a/src/tpm2-tss-engine-common.h
+++ b/src/tpm2-tss-engine-common.h
@@ -33,6 +33,15 @@
 #ifndef TPM2_TSS_ENGINE_COMMON_H
 #define TPM2_TSS_ENGINE_COMMON_H
 
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L \
+    && !defined(__STDC_NO_ATOMICS__)
+# include <stdatomic.h>
+# define TPM2_TSS_ENGINE_HAVE_C11_ATOMICS
+typedef _Atomic int T2TE_ATOMIC_INT;
+#else
+typedef int T2TE_ATOMIC_INT;
+#endif
+
 #include <tpm2-tss-engine.h>
 #include <tss2/tss2_mu.h>
 #include <tss2/tss2_esys.h>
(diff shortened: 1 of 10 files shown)


src/tpm2-tss-engine-common.h
```diff
@@ -33,6 +33,8 @@
 #ifndef TPM2_TSS_ENGINE_COMMON_H
 #define TPM2_TSS_ENGINE_COMMON_H
 
+#include <stdatomic.h>
```
williamcroberts, line 36: I wonder how portable this C11 construct is, any ideas?

src/tpm2-tss-engine-digest-sign.c
```diff
@@ -0,0 +1,304 @@
...
+                                 src_sig_data->seq_handle, &context);
+            ERRchktss(digest_sign_copy, r, goto error);
+            dst_sig_data->seq_handle = ESYS_TR_NONE;
+            r = Esys_ContextLoad(src_sig_data->key->esys_ctx, context,
+                                 &dst_sig_data->seq_handle);
+            ERRchktss(digest_sign_copy, r, goto error);
+        }
+
+        dst_sig_data->key = src_sig_data->key;
+        atomic_fetch_add(&dst_sig_data->key->refcount, 1);
```
williamcroberts, line 258: Since we already have to contextsave/load the sequence handle to copy it for another context, cant we just dup the whole thing and avoid refcounting? Or could we use __atomic_add_fetch since gcc and clang support this since way back? Or perhaps try for atomic.h and if not, fall back to using builtins.

rshearman: My first approach was actually to dup the whole thing and not do the refcounting, but this causes a failure in one of the integration tests due to now running out of transient object handles. I would imagine this isn't a problem in real-world usage of this code as a TPM resource manager will be present, but I don't know for sure and it is annoying for testing purposes.

I can certainly add detection for C11 stdatomic and fall back to the compiler builtins.

src/tpm2-tss-engine-common.h
rshearman: With the detection of C11 atomics and then fallback to gcc/clang builtins then this should cover most cases except for MSVC on Windows. If that needs to be supported then the atomic_fetch_add could be implemented in terms of the native InterlockedExchangeAdd API, or the OpenSSL CRYPTO_atomic_add API.

src/tpm2-tss-engine-digest-sign.c
williamcroberts: Yeah that would be great.

src/tpm2-tss-engine-common.h
williamcroberts: I didn't even realize openessl had an atomic API. It seems that's since OpenSSL_1_1_1, but I could be wrong. Right now configure shows the minimum to be version 1.0.2g, so I'd like to not bump us out of 1.0.2 land, but I could be persuaded.

I like just using OSSL and then we don't have to worry about C11 or compiler support. I don't like that we have to bump the OSSL version

@AndreasFuchsSIT any thoughts here?

williamcroberts (approved):
I'm OK with this, but will leave the final word and merge to @AndreasFuchsSIT

AndreasFuchsSIT:
I am very sorry, but seems like another PR created a merge-conflict with this one.
Could you rebase onto latest master, so I can hit the merge button ?

Also would you be willing to add the same support for RSA keys ?

rshearman:
No worries on the merge conflict.

Sure, I can add the same support for RSA.

AndreasFuchsSIT (approved):
This is some very nice stuff :-)

PR #238, opened 2021-12-06
rand add seed. other improve

aaronUniontech:

### DIFF
--- a/src/tpm2-tss-engine-rand.c
+++ b/src/tpm2-tss-engine-rand.c
@@ -109,10 +109,9 @@ rand_bytes(unsigned char *buf, int num)
         buf += b->size;
     }
 
+ end:
     if(b)
         free(b);
-
- end:
     esys_ctx_free(&esys_ctx);
     return (r == TSS2_RC_SUCCESS);
 }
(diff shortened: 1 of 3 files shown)


src/tpm2-tss-engine-rand.c
```diff
@@ -40,6 +40,46 @@
...
+    ERRchktss(rand_seed, r, goto end);
+
+    TPM2B_SENSITIVE_DATA stir;
+    size_t offset = 0;
+    char* cur_data = (char*)seed;
+
+    static const size_t tpm_random_stir_max_size = 128; 
+    while(offset < (size_t)seed_len) {
+        size_t left = seed_len - offset;
+        // in test, tpm stir seed size beyond 128 will failed with 0x000001d5
```
williamcroberts, line 63: That's per the spec, see 16.2.1 of:
  - https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf

src/tpm2-tss-engine-rand.c
```diff
@@ -67,12 +107,12 @@ rand_bytes(unsigned char *buf, int num)
...
         num -= b->size;
         buf += b->size;
-        free(b);
     }
 
-    esys_ctx_free(&esys_ctx);
-
  end:
+    if(b)
+        free(b);
```
williamcroberts, line 114: if b is null free will ignore it (per the C stdlib spec), so just make sure it's initialized.

aaronUniontech: yes

PR #239, opened 2021-12-08
add rand_seed. rand_bytes reuse buffer b

aaronUniontech:

### DIFF
--- a/src/tpm2-tss-engine-rand.c
+++ b/src/tpm2-tss-engine-rand.c
@@ -55,12 +55,11 @@ rand_seed(const void *seed, int seed_len)
 
     TPM2B_SENSITIVE_DATA stir;
     size_t offset = 0;
-    char* cur_data = (char*)seed;
+    char *cur_data = (char*)seed;
 
     static const size_t tpm_random_stir_max_size = 128; 
     while(offset < (size_t)seed_len) {
         size_t left = seed_len - offset;
-        // in test, tpm stir seed size beyond 128 will failed with 0x000001d5
         size_t chunk = left > tpm_random_stir_max_size ? tpm_random_stir_max_size : left;
 
         stir.size = chunk;
@@ -111,8 +110,7 @@ rand_bytes(unsigned char *buf, int num)
     }
 
  end:
-    if(b)
-        free(b);
+    free(b);
     if(esys_ctx)
         esys_ctx_free(&esys_ctx);
     return (r == TSS2_RC_SUCCESS);
--- a/src/tpm2-tss-engine-rand.c
+++ b/src/tpm2-tss-engine-rand.c
@@ -71,7 +71,7 @@ rand_seed(const void *seed, int seed_len)
             &stir);
         ERRchktss(rand_seed, r, goto end);
 
-        offset += seed_len;
+        offset += chunk;
     }
 
 end:
--- a/src/tpm2-tss-engine-rand.c
+++ b/src/tpm2-tss-engine-rand.c
@@ -40,6 +40,47 @@
 #include "tpm2-tss-engine.h"
 #include "tpm2-tss-engine-common.h"
 
+/** rand seed
+ * @retval 1 on success
+ * @retval 0 on failure
+ */
+static int
+rand_seed(const void *seed, int seed_len)
+{
+    ESYS_CONTEXT *esys_ctx = NULL;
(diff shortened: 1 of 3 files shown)


src/tpm2-tss-engine-rand.c
```diff
@@ -67,12 +108,13 @@ rand_bytes(unsigned char *buf, int num)
         memcpy(buf, &b->buffer, b->size);
         num -= b->size;
         buf += b->size;
-        free(b);
     }
 
-    esys_ctx_free(&esys_ctx);
-
  end:
+    if(b)
```
williamcroberts, line 114: free is skipped if b is null, no need to check.

src/tpm2-tss-engine-rand.c
```diff
@@ -40,6 +40,47 @@
...
+{
+    ESYS_CONTEXT *esys_ctx = NULL;
+    TSS2_RC r;
+
+    r = esys_ctx_init(&esys_ctx);
+    ERRchktss(rand_seed, r, goto end);
+
+    TPM2B_SENSITIVE_DATA stir;
+    size_t offset = 0;
+    char* cur_data = (char*)seed;
```
williamcroberts, line 58: pointer asterisk tight to variable name not type.
`char *cur_data`

src/tpm2-tss-engine-rand.c
```diff
@@ -40,6 +40,47 @@
...
+
+    static const size_t tpm_random_stir_max_size = 128; 
+    while(offset < (size_t)seed_len) {
+        size_t left = seed_len - offset;
+        // in test, tpm stir seed size beyond 128 will failed with 0x000001d5
```
williamcroberts, line 63: This is in the spec part 3 commands for TPM2_StirRandom, so link to it or mention it here over "it fails, not sure why".

williamcroberts (changes requested):

src/tpm2-tss-engine-rand.c
```diff
@@ -40,6 +40,46 @@
...
+        stir.size = chunk;
+        memcpy(stir.buffer, cur_data + offset, chunk);
+
+        r = Esys_StirRandom(
+            esys_ctx,
+            ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE,
+            &stir);
+        ERRchktss(rand_seed, r, goto end);
+
+        offset += seed_len;
```
AndreasFuchsSIT, line 74: Shouldn't this read `offset += chunk` ?

src/tpm2-tss-engine-rand.c
```diff
@@ -67,12 +107,12 @@ rand_bytes(unsigned char *buf, int num)
         memcpy(buf, &b->buffer, b->size);
         num -= b->size;
         buf += b->size;
-        free(b);
```
AndreasFuchsSIT, line 70: This change is incorrect. Esys_GetRandom allocates a new buffer on every invocation.

AndreasFuchsSIT (changes requested):
See my comments please.

src/tpm2-tss-engine-rand.c
aaronUniontech: yes. it's serious.

src/tpm2-tss-engine-rand.c
aaronUniontech: the code in tpm2-tss, master, it's only new buffer while it's null. 

    Esys_GetRandom_Finish

/* Allocate memory for response parameters */
    if (randomBytes != NULL) {
        *randomBytes = calloc(sizeof(TPM2B_DIGEST), 1);
        if (*randomBytes == NULL) {
            return_error(TSS2_ESYS_RC_MEMORY, "Out of memory");
        }
    }

AndreasFuchsSIT:
So all changes to `rand_bytes()` should not be part of this PR, but be a different PR, since they are unrelated to the purpose of StirRandom.
Also you need to squash all your commits into a single one, that includes some more information and also a signed-off tag.
Then, since this is a new functionality, I'd like to see a second commit, that introduces a test script invoking the new functionality.

One we got this, I'll be very happy to merge this.
Sorry for so many requests at once, but that's what makes a good PR. :-)

aaronUniontech:
> 

ok

PR #243, opened 2022-01-04
Using composite actions for CI

malikabhi05:
This PR changes the CI slightly by pointing the builds to common custom composite actions. These actions are currently under malikabh/test-ac and must ideally be moved to their own repo in the tpm2-software org.

This PR is currently for review and discussion and is opened as a draft PR. If everything is good then it can be converted into a proper PR and merged.


AndreasFuchsSIT:
What's the difference of a composite action to what we currently have ?

williamcroberts:
> What's the difference of a composite action to what we currently have ?

The goal is to move all the common stuff out to the other repo so only the non-common stuff remains in the .ci folders for each project.

.github/workflows/main.yml
```diff
@@ -14,13 +14,15 @@ jobs:
...
+      - name: Launch Action
+        uses:
+          malikabhi05/test-ac/runCI@master
+        with:
           DOCKER_IMAGE: "${{ matrix.DOCKER_IMAGE }}"
           TPM2TSS_BRANCH: "${{ matrix.TPM2TSS_BRANCH }}"
           TPM2TOOLS_BRANCH: "${{ matrix.TPM2TOOLS_BRANCH }}"
           CC: "${{ matrix.CC }}"
-        run: ./.ci/ci.sh
+          PROJECT_NAME: "tpm2-tss-engine"
```
williamcroberts, line 25: Could this be "github.event.repository.name" as per: https://stackoverflow.com/questions/58467939/is-there-a-way-to-get-the-project-folder-name-only-on-a-github-action?

malikabhi05: done!

AndreasFuchsSIT:
IMHO you should rebase/sQuash/repack the commits; esp the mention of your private repo should disappear...

williamcroberts:
The action is complete, it's https://github.com/tpm2-software/tpm2-tss-engine/actions/runs/1714787099

So approving and merging.

Issue #245, opened 2022-01-13
Bad free


AndreasFuchsSIT:
These should be `OPENSSL_free()`
https://github.com/tpm2-software/tpm2-tss-engine/blob/c0200843a13fc0d5e7a12e0200ce46859596c1ca/src/tpm2-tss-engine-common.c#L697
https://github.com/tpm2-software/tpm2-tss-engine/blob/c0200843a13fc0d5e7a12e0200ce46859596c1ca/src/tpm2tss-genkey.c#L407
https://github.com/tpm2-software/tpm2-tss-engine/blob/c0200843a13fc0d5e7a12e0200ce46859596c1ca/src/tpm2tss-genkey.c#L411

Also, there are a bunch of `free()` calls that should be `Esys_Free()` all over the project.

PR #246, opened 2022-01-13
Fix allocation scopes

AndreasFuchsSIT:
Use Esys_Free for Esys allocated return parameters.
Use OPENSSL_free for OPENSSL_malloc memory.
Unify to use only OPENSSL_malloc instead of calloc.

Fixes: #245

### DIFF
--- a/src/tpm2-tss-engine-digest-sign.c
+++ b/src/tpm2-tss-engine-digest-sign.c
@@ -172,7 +172,7 @@ digest_sign_init(EVP_PKEY_CTX *ctx, EVP_MD_CTX *mctx, TPM2_DATA *tpm2data,
         /* non-TPM key - nothing to do */
         return 1;
 
-    TPM2_SIG_DATA *data = calloc(sizeof(*data), 1);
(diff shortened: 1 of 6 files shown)


src/tpm2-tss-engine-digest-sign.c
```diff
@@ -268,12 +268,12 @@ digest_sign_copy(EVP_PKEY_CTX *dst, EVP_PKEY_CTX *src)
...
 
-    free(context);
+    Esys_Free(context);
     return 1;
 
  error:
-    free(context);
-    free(dst_sig_data);
+    Esys_Free(context);
+    Esys_Free(dst_sig_data);
```
JuergenReppSIT, line 276: @AndreasFuchsSIT  Shouldn't dst_sig_data freed with OPENSSL_free?

JuergenReppSIT (changes requested):

src/tpm2-tss-engine-digest-sign.c
AndreasFuchsSIT: Correct, fixed.

JuergenReppSIT (approved):

PR #249, opened 2022-02-10
digest-sign: fix digest of large chunks of data

### LINKED ISSUE #248: tpm2-tss-engine:func(151):The provided digest value is too large:src/tpm2-tss-engine-digest-sign.c:115:
With latest code from master, we're observing a failure in the TPM2 TSS engine:

`sudo openssl s_client -connect <hidden>.azure-devices.net:8883 -cert my_cert.pem -key my_tpm_key.tss -keyform engine -engine tpm2tss`

```
<... connects>
140028234097984:error:8009706E:tpm2-tss-engine:func(151):The provided digest value is too large:src/tpm2-tss-engine-digest-sign.c:115:
140028234097984:error:141F0006:SSL routines:tls_construct_cert_verify:EVP lib:../ssl/statem/statem_lib.c:308:
<connection closed>
```

The same call works fine with the engine built from [v1.1.0](https://github.com/tpm2-software/tpm2-tss-engine/releases/tag/v1.1.0).

rshearman:
Fix the use of digest on chunks of data larger than the buffer size
supported by the TPM by iteratively calling Esys_SequenceUpdate on the
chunks until all the data is consumed.

Extend the restricted signing tests to cover this case.

Fixes #248

Signed-off-by: Rob Shearman <rob@graphiant.com>

### DIFF
--- a/src/tpm2-tss-engine-digest-sign.c
+++ b/src/tpm2-tss-engine-digest-sign.c
@@ -106,21 +106,24 @@ digest_update(EVP_MD_CTX *ctx, const void *data, size_t count)
 {
     EVP_PKEY_CTX *pctx = EVP_MD_CTX_pkey_ctx(ctx);
     TPM2_SIG_DATA *sig_data = EVP_PKEY_CTX_get_app_data(pctx);
+    const uint8_t *current_data = data;
     TSS2_RC r;
 
     DBG("digest_update %p %p\n", pctx, ctx);
 
-    TPM2B_MAX_BUFFER digest_data = { .size = count };
-    if (count > sizeof(digest_data.buffer)) {
-        ERR(digest_update, TPM2TSS_R_DIGEST_TOO_LARGE);
-        return 0;
-    }
-    memcpy(&digest_data.buffer[0], data, count);
+    while (count > 0) {
+        TPM2B_MAX_BUFFER digest_data = { .size = count };
+        if (digest_data.size > sizeof(digest_data.buffer))
+            digest_data.size = sizeof(digest_data.buffer);
+        memcpy(&digest_data.buffer[0], current_data, digest_data.size);
+        current_data += digest_data.size;
+        count -= digest_data.size;
 
-    r = Esys_SequenceUpdate(sig_data->key->esys_ctx, sig_data->seq_handle,
-                            ESYS_TR_PASSWORD, ESYS_TR_NONE,
-                            ESYS_TR_NONE, &digest_data);
-    ERRchktss(digest_update, r, return 0);
+        r = Esys_SequenceUpdate(sig_data->key->esys_ctx, sig_data->seq_handle,
+                                ESYS_TR_PASSWORD, ESYS_TR_NONE,
+                                ESYS_TR_NONE, &digest_data);
+        ERRchktss(digest_update, r, return 0);
+    }
 
     return 1;
 }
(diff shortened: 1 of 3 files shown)


src/tpm2-tss-engine-digest-sign.c
```diff
@@ -106,21 +106,24 @@ digest_update(EVP_MD_CTX *ctx, const void *data, size_t count)
...
 
     DBG("digest_update %p %p\n", pctx, ctx);
 
-    TPM2B_MAX_BUFFER digest_data = { .size = count };
-    if (count > sizeof(digest_data.buffer)) {
-        ERR(digest_update, TPM2TSS_R_DIGEST_TOO_LARGE);
-        return 0;
+    while (count > 0) {
+        TPM2B_MAX_BUFFER digest_data = { .size = count };
+        if (digest_data.size > sizeof(digest_data.buffer))
```
AndreasFuchsSIT, line 116: This may not be working, since the C-define for MAX_BUFFER can support up to 64 bytes.
Real world TPMs may however be restricted to 32 bytes for this data structure.
So we should check against 32 bytes here to make sure we run on all hardware TPMs out there.

AndreasFuchsSIT (changes requested):

src/tpm2-tss-engine-digest-sign.c
AndreasFuchsSIT: Ok, I'll take that back, I was mixing this up with TPM2B_DIGEST, where we have this problem.
However I noticed that MAX_BUFFER is at least 1024 bytes (TPM spec part 2 10.4.8). So the test needs to use a longer test value to encounter this case.

williamcroberts: I dunno if the code does any getcap calls and can cache the result of TPM2_PT_NV_BUFFER_MAX from a getcap call for TPM2_CAP_TPM_PROPERTIES. Then you know the value and if that TPM doesn't support that property then default to 32 or whatever.

rshearman: From https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf:
> 17.5 TPM2_SequenceUpdate
> General Description
> This command is used to add data to a hash or HMAC sequence. The amount of data in buffer may be
> any size up to the limits of the TPM.
> NOTE 1 In all TPM, a buffer size of 1,024 octets is allowed

And TPM2B_MAX_BUFFER is:
```c
struct TPM2B_MAX_BUFFER {
    UINT16 size;
    BYTE buffer[TPM2_MAX_DIGEST_BUFFER];
};
```

With TPM2_MAX_DIGEST_BUFFER:
```c
#define TPM2_MAX_DIGEST_BUFFER  1024
```

At least in the version of tpm2-tss that I'm looking at, so I'm not sure I see what the issue is. The tests use 4*512 to exercise this code path (and do indeed fail without the fix to digest_update).

Unless I'm missing something, TPM2_PT_NV_BUFFER_MAX isn't relevant here since that's for non-volatile storage, whereas this is for digest.

williamcroberts: Oh silly me, I thought this was for nv_read, and its not. Disregard. I looked at tpm2_hash.c and tpm2_hmac.c in the tools and we just bound by TPM2_MAX_DIGEST_BUFFER and it seems to work AFAICT. I don't recall getting a bug on it.

AndreasFuchsSIT (approved):

PR #251, opened 2022-04-19
VERSION: add version file

williamcroberts:
Generate the version file with bootstrap and include in the DIST tarball
so endusers can call autoreconf on a dist tarball which doesn't have
git. This alleviates git describe errors on release tarballs in the
autoreconf case.


williamcroberts:
whitespace check did pass: https://github.com/tpm2-software/tpm2-tss-engine/runs/6084064387?check_suite_focus=true

Issue #277, opened 2023-10-09
Provide sm2 algorithm support


bangcheng:
Currently tpm2.0 already supports the use of SM2, SM3, and SM4 algorithms published by State Encryption Management Bureau, China. However, there is still some gap between tpm2-tss-engine supporting these SMx algorithms.

PR #278, opened 2023-10-09
Feature: Provide sm2 algorithm support

bangcheng:
This code is to enable tpm2-tss-engine to support the SMx algorithm. (#277)


src/tpm2-tss-engine-ecc.c
```diff
@@ -796,6 +805,42 @@ tpm2tss_ecc_genkey(EC_KEY *key, TPMI_ECC_CURVE curve, const char *password,
...
+static int pkey_sm2_sign(EVP_PKEY_CTX *ctx, unsigned char *sig, size_t *siglen,
+                         const unsigned char *tbs, size_t tbslen)
+{
+    EC_KEY *ec = EVP_PKEY_get0_EC_KEY(EVP_PKEY_CTX_get0_pkey(ctx));
+
+    if (tpm2tss_ecc_getappdata(ec) == NULL) {
+        // call pkey_sm2_sign_default
+        return pkey_sm2_sign_default(ctx, sig, siglen, tbs, tbslen);
+    }
+    else {
```
kalvdans, line 823: no need for an else clause here, as we already returned if taken the if clause.

kalvdans (approved):
looks neat to me. Haven't tested it though.

PR #279, opened 2023-11-19
Extend key -read and -write functions

chris2511:
If an existing key or the new key is not intended to be stored in a file, the current implementation requires the creation of temporary files with all its odds (cleanup, symlink attack, mktemp)

This change allows to provide the key in many different ways:

A filename of "-" translates to <stdin> for reading and <stdout> for writing.

If the file named by "filename" does not exist,
check whether it contains a "-----BEGIN" and try interpreting the filename itself as key.

Extend documentation in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,15 @@ To export the public key use:
 openssl ec -engine tpm2tss -inform engine -in mykey -pubout -outform pem -out mykey.pub
 ```
 
+## The TPM Key
+The TPM key may be provided as file name, like "mykey" in the examples above.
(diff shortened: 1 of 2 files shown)


AndreasFuchsTPM (approved):

TaeZStkyoht:
@chris2511 This is really useful and we were also thinking in the same way. I was using /tmp folder for key creation while using tpm2tss-genkey but our static code analyzer Sonar was warning about this security hotspot.
This functionality definitely must be on this library.

PR #285, opened 2024-09-19
Makefile.am: Fix build without pandoc

### LINKED ISSUE #284: tarball: building without pandoc fails after `make clean; configure`
When using the source tarball, and building without pandoc, builds fail after `make clean` with a subsequent re-`configure`.

This is what happens when building multiple times in a Yocto workspace, e.g. using this recipe in the meta-security layer:
https://git.yoctoproject.org/meta-security/tree/meta-tpm/recipes-tpm2/tpm2-tss-engine/tpm2-tss-engine_1.1.0.bb?h=kirkstone&id=353078bc06c8b471736daab6ed193e30d533d1f1

When re-using the workspace. Yocto will issue a `make clean` before running `autoreconf`, `./configure` and the building.

Since the tarball includes the pre-processed man pages under `man/man{1,3}/`, and make clean erases the man page files (but not the directory), the subsequent `./configure` picks up `HAVE_MAN_PAGES`:
https://github.com/tpm2-software/tpm2-tss-engine/blob/766505bf5c943c614fd246d27d1e5cd66543250b/configure.ac#L139
and tries to build them, even when pandoc is not available. This results in a build failure:
```
  GEN      man/man1/tpm2tss-genkey.1
/bin/sh: line 1: -s: command not found
```
from this Makefile line:
https://github.com/tpm2-software/tpm2-tss-engine/blob/766505bf5c943c614fd246d27d1e5cd66543250b/Makefile.am#L205-L206

The question is whether the pre-processed man pages just should not be deleted if they cannot be built, or whether the  `HAVE_MAN_PAGES` condition should be determined differently.

barsnick:
When building from a tarball, pre-created man pages are preset. Yet `make clean` clobbers them, leaving their target directory intact. Therefore, the next configure call will try to build man pages even if pandoc isn't available.

Fix this by cleaning the man pages only if they can be recreated by the presence of pandoc.

Fixes https://github.com/tpm2-software/tpm2-tss-engine/issues/284

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -217,8 +217,10 @@ EXTRA_DIST += \
     man/tpm2tss_ecc_genkey.3.md \
     man/tpm2tss_ecc_getappdata.3.md
 
+if HAVE_PANDOC
 CLEANFILES += \
     $(dist_man_MANS)
+endif
 
 ### Bash Completion
 bash_completiondir = $(completionsdir)


AndreasFuchsTPM (changes requested):
Please add a Signed-of-by tag

barsnick:
> Please add a Signed-of-by tag

I have no idea why the DCO check fails to see it. It's been there all along.

barsnick:
> I have no idea why the DCO check fails to see it.

Ah, I got it. It needs to be identical to the author.

PR #294, opened 2025-04-03
fix memory leaks

TaeZStkyoht:
- BIGNUM must be freed both on tpm2tss_tpm2data_write and tpm2tss_tpm2data_read:
definitely lost: 24 bytes in 1 blocks
indirectly lost: 8 bytes in 1 blocks
- TSSPRIVKEY's parent, privkey and pubkey members don't need to be allocated explicitly, because TSSPRIVKEY_new already does:
definitely lost: 72 bytes in 3 blocks
indirectly lost: 0 bytes in 0 blocks
- TPM2_DATA* returned from RSA_get_app_data can be returned directly genkey_rsa() because RSA_free() doesn't free this item:
definitely lost: 2,248 bytes in 1 blocks
indirectly lost: 0 bytes in 0 blocks
- update .gitignore to cover more generated files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 Makefile
 Makefile.in
 aclocal.m4
+ax_*.m4
+pkg.m4
 aminclude_static.am
 autom4te.cache/
 compile
@@ -16,6 +18,7 @@ install-sh
 libtool
 libtpm2.la
 ltmain.sh
+stamp-h1
 m4/libtool.m4
 m4/ltoptions.m4
 m4/ltsugar.m4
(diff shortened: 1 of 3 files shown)


kalvdans (approved):

williamcroberts:
You have four different tasks that were mentioned in the commit message, one for each of the list items. I think this should be broken up into four commits instead of one, so please break this up into four commits.

TaeZStkyoht:
> You have four different tasks that were mentioned in the commit message, one for each of the list items. I think this should be broken up into four commits instead of one, so please break this up into four commits.

done

williamcroberts (approved):

kalvdans (approved):

TaeZStkyoht:
I don't know about those errors on pipelines.

williamcroberts:
> I don't know about those errors on pipelines.

I don't think it's you, I need to fix this

Issue #270, opened 2023-03-29
TPM engine couldn't load when loading the tpm engine more than once with 1.2.0 version of TPM2-TSS-ENGINE


shrikant1407:
we are trying to load the TPM engine using OpenSSL API as follows:

``` 
ENGINE_load_dynamic();
engine = ENGINE_by_id("dynamic");
if (engine == NULL) {
	LOG(LOG_ERROR, "Could not find external engine.\n");
	goto error;
}

if (!ENGINE_ctrl_cmd_string(engine, "SO_PATH", TPM2_TSS_ENGINE_SO_PATH,
				    0)) {
	LOG(LOG_ERROR, "Could not set TPM Engine path.\n");
	goto error;
}

if (!ENGINE_ctrl_cmd_string(engine, "LOAD", NULL, 0)) {
	LOG(LOG_ERROR, "Could not load TPM engine.\n");
	goto error;
}
```
The flow for loading such an engine is triggered by `ENGINE_ctrl_cmd_string(engine, "SO_PATH", TPM2_TSS_ENGINE_SO_PATH,
				    0)`, and when using a dynamic path it will dynamically load the engine's `.so`file. Our application needs to load the TPM engine more than once but it's not able to load at the second instance. The same configuration was working with **1.1.0** version of TPM2-TSS-ENGINE

AndreasFuchsTPM:
Could you provide me with a complete snippet that we can add to the tests ?

PerryWerneck:
I think I'm got the same problem in wpa_supplicant.

From the wpa_supplicant log:

At the beggining...

```
1750771446.519763: TLS: using phase1 config options
1750771446.520722: OpenSSL: Found TPM2 wrapped key /usr/share/cdsn/hom/1286043559000.key
1750771446.520763: SSL: Initializing TLS engine tpm2
1750771446.544986: ENGINE: engine initialized
1750771446.545771: TLS: Trusted root certificate(s) loaded
1750771446.548131: OpenSSL: SSL_use_certificate_chain_file --> OK
1750771446.548171: TLS: Using private key from engine
1750771446.548272: eth0: CTRL-EVENT-EAP-METHOD EAP vendor 0 method 13 (TLS) selected
1750771446.548309: EAP: EAP entering state METHOD
1750771446.548345: SSL: Received packet(len=6) - Flags 0x20
```

But, later in the same log...


```
[...]

PR #299, opened 2026-03-26
fix: use EVP_DigestFinal_ex in rsa_signctx for OpenSSL 3.x compatibility

mariosergiosl:
In OpenSSL 3.x, EVP_MD_CTX_set_update_fn() is deprecated and silently ignored, so digest_update() is never called and digest_finish() returns SHA-256 of empty string. Replace digest_finish() with EVP_DigestFinal_ex() for unrestricted keys to pass the correct hash to Esys_Sign().

Fixes EAP-TLS authentication failure with wpa_supplicant and NetworkManager. Verified on openSUSE Leap 15.6 (OpenSSL 3.1.4) and SLES 16 (OpenSSL 3.5.0).

[PR_DESCRIPTION.md](https://github.com/user-attachments/files/26287656/PR_DESCRIPTION.md)

### DIFF
--- a/src/tpm2-tss-engine-rsa.c
+++ b/src/tpm2-tss-engine-rsa.c
@@ -717,12 +717,31 @@ rsa_signctx(EVP_PKEY_CTX *ctx, unsigned char *sig, size_t *siglen,
         return 0;
     }
 
-    if (!digest_finish(sig_data, &digest_ptr, &validation_ptr))
-        return 0;
-
-    r = Esys_Sign(sig_data->key->esys_ctx, sig_data->key->key_handle,
-                  ESYS_TR_PASSWORD, ESYS_TR_NONE, ESYS_TR_NONE,
-                  digest_ptr, &in_scheme, validation_ptr, &tpm_sig);
+    /* OpenSSL 3.x does not honour EVP_MD_CTX_set_update_fn, so the TPM
+     * sequence (Esys_SequenceUpdate/Complete) never receives data and
+     * digest_finish returns SHA-256 of empty string. For unrestricted
+     * keys we finalize the digest via EVP_DigestFinal_ex and pass it
+     * directly to Esys_Sign with a NULL validation ticket.
+     */
+    {
+        unsigned char md_buf[EVP_MAX_MD_SIZE];
+        unsigned int md_len = 0;
+        TPM2B_DIGEST local_digest = { .size = 0 };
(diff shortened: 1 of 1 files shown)


kalvdans:
Where in the documentation does OpenSSL say that the function is silently ignored? Deprecation is one thing, removing functionality is another thing.

mariosergiosl:
Good point — and fair to ask for documentation.

You're right that deprecated doesn't mean removed, and the function call
itself still works. The issue is behavioral: in OpenSSL 3.x, when a digest
is initialized via the provider model (the default since 3.0), the custom
update function registered with EVP_MD_CTX_set_update_fn() is not called
by EVP_DigestUpdate(). The OpenSSL team described exactly this when they
deprecated it — see openssl/openssl#14003:

  "These functions do not fit with the provider model where the update
   function is not used. It is called only from the legacy code."

We confirmed this empirically. With debug logging enabled, digest_update()
never appears in the call sequence during EVP_DigestSignFinal(). The engine
goes directly from key load to rsa_signctx, skipping the update entirely:

  ENGINE: engine initialized
  TLS: Using private key from engine
  rsa_signctx ... sig_data 0x...   ← no digest_update before this

As a result, Esys_SequenceComplete receives no data and returns SHA-256 of
an empty string, which is the hash the TPM then signs:

  e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855

Signature verification then fails with:
  error:02000068:rsa routines:ossl_rsa_verify:bad signature

This was reproduced on OpenSSL 3.1.4 and 3.5.0. OpenSSL 1.x is not
affected because it does not use the provider path.

Also worth noting: Perry Werneck (PerryWerneck/tpmtest  https://github.com/PerryWerneck/tpmtest), who originally
scoped this problem for SUSE engineering, observed the same engine
initialization failure in wpa_supplicant in issue #270 — further
confirming the behavior in real EAP-TLS scenarios.

Full debug logs and lab documentation are available at
https://github.com/mariosergiosl/8021x-eap-tls-ibmtss if useful.

Happy to provide the full debug log or any additional detail if helpful.

williamcroberts:
This is one of those things where I consider, do we even bother to fix. If you're using the engine, and you're on OpenSSL 3.X, switch to provider. Once we take the patch, we would have to cut another release of this project, and eventually this will get mothballed as OpenSSL marches forward.

PR #301, opened 2026-07-03
bug: fix const correctness bugs

williamcroberts:
Some functions, including callbacks, had const in them, while not quite right, they were not quite wrong either, but older OpenSSL was notoriously full of places where const should have been used. So, to keep the patch simple and so we don't have to cast function pointers, drop const.


williamcroberts:
Ok this green lights our CI and patch flow again, so I am going to go ahead and merge this.

